Issue #13, opened 2020-10-24
Click clusters on global map to zoom in


xlcrr:
On supercluster.vue, it would be nice if the user was able to zoom in by clicking a cluster

PR #43, opened 2020-11-29
feature/click-cluster-on-global-map-to-zoom-in

lhtrung307:
Add event listener when user click to cluster.
Please help me to verify if this PR is enough for #13


xlcrr:
That works, thanks!

PR #45, opened 2020-12-01
Apply eslint, fix alignment signup, add smooth wheel zoom, change tag style, change about page sentence and fix mispelling

lhtrung307:
I've rebased some commit that I commit with the wrong email, please help me to check if app.js is right. Sorry for this inconvenience.

PR #46, opened 2020-12-02
Bugfix/picture close automatically if click green dot on top screen

lhtrung307:
Fix picture closes automatically if you open a green dot that is on the top of the screen

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,6 +29,7 @@
         "max-len": ["error", { "code": 120, "ignoreUrls": true, "ignoreComments": true }],
         "vue/html-indent": ["error", 4],
         "vue/max-attributes-per-line": "off",
-        "quotes": ["error", "single"]
+        "quotes": ["error", "single"],
+        "brace-style": ["error", "allman"]
     }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -5397,6 +5397,15 @@
                         "resolve-from": "^4.0.0"
                     }
                 },
+                "lru-cache": {
+                    "version": "6.0.0",
+                    "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz",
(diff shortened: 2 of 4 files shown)


resources/js/views/Locations/Countries.vue
```diff
@@ -61,13 +61,13 @@
                     <div class="columns">
                         <div class="column is-half is-offset-3">
                             <div class="columns">
-                                <div class="column is-one-third">
-                                    <h1 class="subtitle is-5" style="color: black; text-align: center;">
```
xlcrr, line 65: Why are we removing the black here?

resources/js/views/Locations/Countries.vue
```diff
@@ -61,13 +61,13 @@
...
+                                <div class="column has-text-centered">
+                                    <h1 class="subtitle is-6">
                                         <strong style="color:black;">
                                             {{ $t('location.maps10') }}
                                         </strong>
                                     </h1>
-                                    <h1 class="title is-2" style="text-align: center;">
+                                    <h1 class="title is-2">
```
xlcrr, line 70: Why are we removing `center` here?

xlcrr (changes requested):
A lot of similarities here between this and #45  

What does this PR fix exactly ?

resources/js/views/Locations/Countries.vue
lhtrung307: We already have `style="color:black;"` on the `strong` tag.

resources/js/views/Locations/Countries.vue
lhtrung307: We already have `has-text-centered` on the `div` tag.

lhtrung307:
Like I said in slack, this for `Fix picture closes automatically if you open a green dot that is on the top of the screen`. And you should merge it after PR#45. I want to use the eslint so I created this branch base on that.

xlcrr (approved):

PR #49, opened 2020-12-05
Change team UI, handle upload error

lhtrung307:


xlcrr:
Unfortunately I don't think the side-nav is going to work for mobile. We should consider using top-nav

xlcrr:
Top nav on production

PR #70, opened 2020-12-06
Add Settings to Teams

xlcrr:
We want to give the user to control privacy for each team they have joined


resources/js/views/Teams/TeamSettings.vue
```diff
@@ -0,0 +1,152 @@
...
+
+<style scoped>
+
+    .tsc {
+        margin-top: 1em;
+        margin-left: 5em;
+    }
+
+
+    @media screen and (max-width: 768px)
```
lhtrung307, line 144: Try using the `breakpoints` I've defined.

xlcrr: I will, but I was in a rush

PR #76, opened 2020-12-09
feature/add-user-avatar-to-personal-settings

lhtrung307:
Allow users to change their avatar


xlcrr:
Thanks! I think we should keep these assets on S3 instead of our public folder

lhtrung307:
Yeah, I just follow the old code and also don't have access to S3 😄. Really happy if you help me fix the code upload to S3.

xlcrr:
I got started on this last night but its not needed yet

We will introduce profile.vue later with profile picture that users can upload if they want 

The code to upload images is available on PhotosController.php @ store

PR #88, opened 2020-12-14
Feature/bounding box

willalexander:
Implemented individual box selection & deletion

Created a const value for minimum box size

### DIFF
--- a/resources/js/views/admin/BoundingBox.vue
+++ b/resources/js/views/admin/BoundingBox.vue
@@ -99,7 +99,8 @@ export default {
          */
         image ()
         {
-            return 'backgroundImage: url(' + this.$store.state.admin.filename + ')';
+            //return 'backgroundImage: url(' + this.$store.state.admin.filename + ')';
+            return 'backgroundImage: url(assets/plastic_bottles.jpg)';
         },
 
         /**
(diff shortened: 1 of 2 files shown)


resources/js/views/admin/BoundingBox.vue
```diff
@@ -216,7 +218,10 @@ export default {
         {
             if(this.activatedBox == -1)
             {
+                // Record the box which has been selected, and set all other boxes to be deselected:
+                for(var b = 0; b < this.boxes.length; b++) this.boxes[b].selected = false;
```
lhtrung307, line 222: Should use forEach for shorter code and readability 😄.

resources/js/views/admin/BoundingBox.vue
```diff
@@ -226,8 +231,12 @@ export default {
         /**
          *
          */
-        startDrawingBox (e)
+        mousedownSelf (e)
         {
+            // The user has clicked outside all the boxes, so deselect all boxes:
+            for(var b = 0; b < this.boxes.length; b++) this.boxes[b].selected = false;
```
lhtrung307, line 237: Create a function to reuse instead of duplicate code with line 222.

resources/js/views/admin/BoundingBox.vue
willalexander: thanks, done!

resources/js/views/admin/BoundingBox.vue
willalexander: done!

xlcrr (approved):

PR #89, opened 2020-12-15
feature/improve-ui-for-user-settings

lhtrung307:
Update user setting sidebar, combine personal details, email subscription, presence to 1 menu. Update mobile and desktop UI for each setting.

olm test_settings_details

olm test_settings_details (1)

olm test_settings_details (4)


xlcrr:
Thanks but we will not merge this 

First- we should be focusing on critical bugs that are blocking users. Very few people use these settings. If it's not broken, no need to fix it yet.

Second- I don't want too many things on each page. Keep it simple

PR #90, opened 2020-12-15
Generate dynamic translation string for all tags

xlcrr:
The way we were generating tags for display was cumbersome and static.

Now we can generate all translation key => value pairs for all tags in any image in 1-line of code

Some improvements could be made to the frontend translation which executes onEachFeature as the global-points are displayed. For a lot of features, this could get slow. The translation should only happen for each feature when it is clicked and opened.


resources/js/views/global/Supercluster.vue
```diff
@@ -64,22 +65,36 @@ function createClusterIcon (feature, latlng)
...
  */
 function onEachFeature (feature, layer)
 {
     if (! feature.properties.cluster)
     {
+        let a = feature.properties.result_string.split(',');
+        a.pop();
+
+        let z = '';
+        a.forEach(i => {
```
lhtrung307, line 81: Should use `reduce` instead of `forEach`

resources/js/views/global/Supercluster.vue
```diff
@@ -115,38 +128,34 @@ async function update ()
         await axios.get('clusters', {
             params: { zoom, bbox }
         })
-            .then(response =>
-            {
-                console.log('get_clusters.update', response);
+        .then(response => {
```
lhtrung307, line 131: No consistency between function bracelet and arrow function bracelet. Should define one style and follow it.

resources/js/views/global/Supercluster.vue
xlcrr: We shouldn't loop here. The translation should only happen when the image is clicked and loaded.

resources/js/views/global/Supercluster.vue
xlcrr: For requests, I use `=> {` but for `ifs` and `else` they go on a new line. Happy to keep everything the same if it helps

PR #91, opened 2020-12-16
feature/handle-error-when-upload-litter-picture

lhtrung307:
Handle error when upload litter picture.

### DIFF
--- a/resources/js/views/general/Upload.vue
+++ b/resources/js/views/general/Upload.vue
@@ -80,7 +80,7 @@ export default {
 
             if(file.xhr?.response && file.xhr?.response?.length < 200)
             {
-                errorMessage = file.xhr?.response;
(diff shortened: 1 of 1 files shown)


resources/js/views/general/Upload.vue
```diff
@@ -69,8 +75,19 @@ export default {
         verror (file)
         {
             console.log({ file });
-            // this.error = true
-            // window.toastr.error(file.upload.filename, 'Event : vdropzone-error - ' + file.status)
+
+            let errorMessage = 'Something went wrong. Please try again later.';
```
xlcrr, line 79: We need to be more descriptive and it should be translated

When an image does not have a geotag, the error currently says "Sorry, no GPS tag found". This error would overwrite that?

lhtrung307: Yes, this error just in case we got some unknown error.

PR #102, opened 2020-12-18
feature/improve-ui-for-about-page

lhtrung307:
Improve mobile and desktop UI for about page

### DIFF
--- a/public/css/openlittermap.css
+++ b/public/css/openlittermap.css
@@ -2,10 +2,6 @@
 	visibility: hidden;
 }
 
-html {
-    height: 100vh;
-}
-
 /* Extra leaflet styles */
 .info {
     padding: 6px 8px;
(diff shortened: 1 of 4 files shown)


xlcrr:
Can you provide screenshots? Also that 100vh is a global setting. Did you check for unintended consequences?

lhtrung307:
It's a long page, you should check it. And I've checked for 100vh.

xlcrr:
Can anyone else take a look at this? It would help

xlcrr:
Doesn't look good for me. Am I missing something?

Screenshot 2021-01-17 at 15 45 13

xlcrr:
Screenshot 2021-01-17 at 15 46 06

This should also be positioned absolutely on the bottom-right of the map, and it should stay there as we change the size of the browser window

PR #104, opened 2020-12-19
Fred/issue93attempt2

FredMark:
I improved the column-titles and added a lot of columns that users could use in app/web,
But where not shown in the export.

There are still a few tasks to do, but i will create a new issue for those because there were a different kind of issue and i do not want to solve these in this PR. Better to seperate them.


xlcrr (approved):

xlcrr:
Duplicated by #105

Issue #83, opened 2020-12-12
Number of Images left to tag not correct


FredMark:
I have uploaded some pictures and i am ready to start tagging the first one.
In the top right, it says 
'Images left to tag: 1'

In the bottom there is a button 'Next image' and 17 squares.

The number in the top right is not correct, there is not 1 image left to tag, but 17.
![Schermafdruk 2020-12-11_20 28 32 (irfanview)](https://user-images.githubusercontent.com/2518047/101984387-7a958880-3c81-11eb-8820-c0fe1119c7bc.jpg)

PR #110, opened 2020-12-23
bugfix/number-of-image-left-to-tag-not-correct

lhtrung307:
Fix number of images left to tag not correct. Issue #83


app/Http/Controllers/PhotosController.php
```diff
@@ -400,18 +400,16 @@ public function unverified ()
...
             'user_id' => $user->id,
             'verified' => 0,
             'verification' => 0
-        ])->count();
+        ];
+
+        $photos = Photo::select('id', 'filename', 'lat', 'lon', 'model', 'remaining', 'display_name', 'datetime')
+            ->where($query)->simplePaginate(1);
+
+        $remaining = Photo::where($query)->count();
```
xlcrr, line 412: Can you change this to

```
$query = Photo::where([
            'user_id' => $user->id,
            'verified' => 0,
            'verification' => 0
]);

$photos = $query->select('id', 'filename', 'lat', 'lon', 'model', 'remaining', 'display_name', 'datetime')->simplePaginate(1);

$remaining = $query->count();
```

resources/js/views/general/Tag.vue
```diff
@@ -21,8 +19,7 @@
 
             <!-- Image is available for tagging -->
             <div v-else>
-
-                <div v-for="photo in photos" class="mb2">
+                <div v-for="photo in photos" :key="photo.id + photo.filename" class="mb2">
```
xlcrr, line 22: `:key="photo.id"` is enough

resources/js/views/general/Tag.vue
```diff
@@ -223,23 +220,27 @@ export default {
          */
         async confirmDelete (photoid)
         {
-            if (confirm("Do you want to delete this image? This cannot be undone."))
+            if (confirm('Do you want to delete this image? This cannot be undone.'))
```
xlcrr, line 223: Needs translation

xlcrr (changes requested):
1-2 small improvements and it can be merged. Thanks

resources/js/views/general/Tag.vue
lhtrung307: `photo.id` is just a number like an index, we should not use an index as a key.

resources/js/views/general/Tag.vue
lhtrung307: Updated. Thanks

app/Http/Controllers/PhotosController.php
lhtrung307: I think you should not do that because it will get all data and then select and pagination. That will create a performance issue.

resources/js/views/general/Tag.vue
xlcrr: Actually, photo.id is unique key. It is not `photos.map( (item, **index**) =>`

resources/js/views/general/Tag.vue
xlcrr: It actually needs a translation key, like `{{ $t('tag.confirm-delete') }}` and we need to add the translation key to each language folder. We also have a new excel sheet to track translation keys

app/Http/Controllers/PhotosController.php
xlcrr: I understand your concern but thankfully it won't get a large data object. 1. `$query` will simply create the query without getting any data. 2. `->simplePaginate(1);` will get 1 object, that has an array of 1 photos. 3. we can use the same `$query` to count the table rows, without actually loading all the data

xlcrr:
Sorry for taking so long to get around to this. We are back to web now

resources/js/views/general/Tag.vue
lhtrung307: Yeah, I've added the translation key, did you check the new commit? I will add translation key to each language folder.

app/Http/Controllers/PhotosController.php
lhtrung307: Updated. Thanks

resources/js/views/general/Tag.vue
lhtrung307: Updated. But you should read more about key in Vue.

PR #115, opened 2020-12-24
Added category Dumping and Industrial where it was missing in php-files

FredMark:
Searched all php-files for 'Sanitary' and looked within that file how it was working for sanitary and checked if the same was happening for 'Dumping' and 'Industrial' that seemed to be missing in several files.
Where it was missing, i added it.


xlcrr:
@FredMark Unfortunately a lot of these files are not used anymore which is why they seem out of date. Appreciate the help but cannot merge this. Need to clean up the existing files.

PR #234, opened 2021-06-06
Bug/broken migrations init setup

mdougllas:
Fixing some minor issues with migrations. This will prevent that new local setups brake when migrating tables for the first time.


xlcrr:
Some of these older column names are still present in the code but many of these are older files that can be deleted

Issue #116, opened 2020-12-26
Donate page, 'updating....brb'?


FredMark:
Go to the homepage,
go to the footer,
click on 'Single Donation'.

See printscreen with strange text 'loading....brb'.
![Schermafdruk 2020-12-26_14 12 03 (irfanview)](https://user-images.githubusercontent.com/2518047/103152120-172d4f80-4785-11eb-9e57-a7fcb35232d2.jpg)
![Schermafdruk 2020-12-26_14 12 12 (irfanview)](https://user-images.githubusercontent.com/2518047/103152123-1bf20380-4785-11eb-8534-9b20d9125718.jpg)

PR #238, opened 2021-06-07
Fixed Donations page loading text & small UI tweaks, seed donations for new setups (#116)

GeniJaho:
- Haven't compiled assets because I'm on Windows and that has sometimes caused me problems
- Not sure about the Vue implementation with loading state. It's straightforward and works, but there might be a better way to do it.

Put everything in a container div 
image

And prevented the image on mobile from stretching. 
image

### DIFF
--- a/resources/js/components/DonateButtons.vue
+++ b/resources/js/components/DonateButtons.vue
@@ -46,8 +46,12 @@ export default {
 	async created ()
 	{
 		this.loading = true;
-		await this.$store.dispatch('GET_DONATION_AMOUNTS');
-		this.loading = false;
+		await this.$store.dispatch('GET_DONATION_AMOUNTS')
+            .then(() => {
+                this.loading = false;
+
+                this.$emit('donations-loaded')
+            });
 	},
 
 	computed: {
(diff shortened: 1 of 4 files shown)


resources/js/components/DonateButtons.vue
```diff
@@ -46,8 +46,12 @@ export default {
...
 	{
 		this.loading = true;
-		await this.$store.dispatch('GET_DONATION_AMOUNTS');
-		this.loading = false;
+		await this.$store.dispatch('GET_DONATION_AMOUNTS')
+            .then(() => {
+                this.loading = false;
+
+                this.$emit('donations-loaded')
+            });
```
xlcrr, line 54: No need to call `then` here, it won't run until after await anyway

xlcrr (approved):
Need to test this before merge

resources/js/components/DonateButtons.vue
GeniJaho: True, missed that. Added a commit for it.

GeniJaho:
Any update on this one? I can remove the Vue changes if they feel to risky.

PR #239, opened 2021-06-10
Initial tests suite & factories

GeniJaho:
- Fixed PHPUnit config file (can't use SQLite)
- Tests use a MySQL database called **olm_test** on the same DB connection
- By default, all tests use the RefreshDatabase trait, included in the base test class
- Added .idea to .gitignore, those files keep showing in diffs
- Unit tested the Photo model

I noticed that Coastal and Art models need a photo_id to be created, but the others don't. Is that ok, or they shouldn't need it either?
Also, the Photo model has Country, State, City both as relationships and DB fields. The relationships are wrong and they don't work, instead of **hasOne** they should be **belongsTo**. Are these relationships used elsewhere in the project? Can we delete them?


xlcrr:
Thanks!

photo_id can be removed from Coastal and Art. Storing the 1-way relationship from photo->$category is enough

I think we can delete them

app/Models/Photo.php
```diff
@@ -173,13 +176,8 @@ public function total ()
...
             {
                 // We dont want to include brands in total_litter
-                if ($category === 'brands')
-                {
-                    $this->$category->total();
-                }
-
                 // Increment total_litter when its not brands
-                else
+                if ($category !== 'brands')
```
xlcrr, line 180: We still need to call the first block, which calculates and saves `total_brands` on the brands model.

The second block increments `total_litter`, which is a sum of all litter, expect brands

We could brands and litter separately which is why this block is needed here

database/factories/Litter/Categories/AlcoholFactory.php
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Define the model's default state.
+     *
+     * @return array
+     */
+    public function definition()
+    {
+        return [
+            'beerCan' => $this->faker->randomDigit
```
xlcrr, line 25: You can do `$this->$category->types()` on any money to get a list of keys

tests/Unit/Models/PhotoTest.php
```diff
@@ -0,0 +1,308 @@
...
+            'brands_id' => $brands->id
+        ]);
+
+        // Suppress output to console
+        $this->setOutputCallback(function() {});
+
+        $photo->total();
+
+        // Brands are not calculated
+        $this->assertEquals($smoking->total(), $photo->total_litter);
```
xlcrr, line 103: We should count total_brands separately

xlcrr (changes requested):

app/Models/Photo.php
GeniJaho: Actually $this->$category->total() does not do anything on the LitterCategory models, it doesn't increase their counter. Or am I missing something?
Anyway I'm setting the method's signature as it was before.

database/factories/Litter/Categories/AlcoholFactory.php
GeniJaho: I actually wanted to keep the factories separate from app logic. How can I use $this->$category->types() in a factory? The types() method is not static, and we don't have a model inside the factory?

database/factories/PhotoFactory.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Define the model's default state.
+     *
+     * @return array
+     */
+    public function definition()
+    {
+        return [
+            'user_id' => User::factory()->create(),
```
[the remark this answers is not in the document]
GeniJaho: We don't need to provide an id. Actually just checked the docs and we can go a step further 😂 
```
    return [
        'user_id' => User::factory(),
        'title' => $this->faker->title(),
        'content' => $this->faker->paragraph(),
    ];
```

tests/Unit/Models/PhotoTest.php
```diff
@@ -0,0 +1,308 @@
...
+        $photo = Photo::factory()->create();
+
+        $annotation = Annotation::factory()->create([
+            'photo_id' => $photo->id
+        ]);
+
+        $this->assertInstanceOf(Collection::class, $photo->boxes);
+        $this->assertCount(1, $photo->boxes);
+        $this->assertTrue($annotation->is($photo->boxes->first()));
+    }
```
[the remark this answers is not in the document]
GeniJaho: I don't quite understand. The relationship is a hasMany, that's why I called it _has_many_boxes_. True, I only pass it only one box, but I assert that it's a Collection, not an Annotation instance.

How do we test for only one box? You mean we add a new box() relationship on the model?

app/Models/Photo.php
xlcrr: Actually you are right, this line can be deleted. `total()` returns a total count of all items on that modal, which is used in some places but not here. Your original code was correct and can be accepted

database/factories/Litter/Categories/AlcoholFactory.php
xlcrr: Leave it for now, we will look into making them static in another PR

tests/Unit/Models/PhotoTest.php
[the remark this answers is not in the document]
xlcrr: I was thinking we could do one test for `assertCount(1)` and another one for `assertCount(n)` but this is okay for now

PR #248, opened 2021-06-17
Test coverage - PhotosController

GeniJaho:
This PR focuses on the Photo Upload feature. Aside from tests, I've modified the PhotosController because:

- The `header()` calls didn't go well in a testing env, they kept throwing "headers already sent" errors, so I converted those into `abort()` calls instead.
- Added imports for `Auth `and `Log `facades, and the `\Exception` class, cleaning it up a little.
- There was a bug when saving images on development. Before they were being saved on `public/local-uploads/2021/06/15/15oifngbdfoigb.jpg`. The day was being appended to the image hash name, and the image was stored in the month directory. Adding the missing slash fixes both the image location and its pathname. This change does not affect existing images, only the newly added ones.

This does not test for images containing `DateTimeOriginal`, `DateTime`, or `Model` Exif attributes. Couldn't find images to fulfill those conditions.

Since photos are uploaded to the public directory on dev environments, it was difficult to use the usual Storage fake to upload images. Instead, I uploaded a test image, which contains geolocation data, and deleted it at the end of the tests.

To mimic production environments we need to use a call to `app()->detectEnvironment()`, which sets the testing env as production. This, in turn, will verify post requests against CSRF tokens and make them fail, because usually, CSRF checking is disabled during testing. To fix that, we need to use the `WithoutMiddleware` trait. Using that trait on the `UploadPhotoTest` caused the authentication test to fail, so I had to port it to another test class `UploadPhotoToProductionTest`.

Ideally, we'd have the same logic applied to both the production and dev envs, and change only the storage driver used in our local/prod .env files. Maybe I can make a PR for that later.

Notice that even for the production environment upload tests we fake the Storage, which means uploading happens only locally. It would be nice to have another test that actually uploads to S3, and then deletes the image at the end of the test, just to be sure that it really works.


xlcrr (approved):
Incredible PR, thank you!!

xlcrr:
I need to test that change to aborting the photo upload, after that will merge

xlcrr:
Screenshot 2021-06-19 at 14 28 42

We need to show the error to the user

PR #261, opened 2021-07-12
Api docs

GeniJaho:
Have used this package to generate api docs: https://github.com/DarkaOnLine/L5-Swagger

I've also run a `composer update`, since the package above needed some updated dependencies.

The docs are not perfect nor complete, but I'll provide updates in a follow-up PR, as I learn to make the docs even more detailed. Meanwhile, we can get feedback from the front-end side if the docs are not clear.


xlcrr:
I'm not a fan of these huge comment blocks. Do they get created automatically or do we have to write them? I don't think we will use this package or at least these huge ugly comment blocks

GeniJaho:
We have to write them manually, unfortunately. You can't get the right level of detail from the existing code.
This package is just a wrapper of the Swagger-php for Laravel, which follows OpenApi standards. I can look for other packages, but I believe this is the best for API docs. I think it's generally accepted to document methods like this, because of two benefits: first, you have the docs in the same place as your code, and secondly, front-end devs can access the docs from a ready-made, widely accepted UI, like this:
image


However, I think we can use separate JSON or YAML files to document the routes without resorting to those long comments. I'll see if we can do that, and will let you know.

GeniJaho:
After carefully checking the docs of the Swagger package, the only way we can generate docs is by using Annotations in code, which in turn generate a JSON OpenApi specification.

If we're going forward with this package, one solution might be to lower the level of detail in our annotations, just keep the basic info needed for the API users. Also, we don't need these comments for the whole codebase, only for 3-4 API controllers.

The two other less-used packages I found were https://github.com/mpociot/laravel-apidoc-generator and https://github.com/knuckleswtf/scribe. The first one uses the same approach with Annotations, so I don't see any benefit in switching to it. The second generates the docs from the existing codebase, but I think we won't get good enough results since the API controller code doesn't use request validation, object types, etc., and we'd still need to complement a lot of documentation by writing it manually.

Let me know what you think.

xlcrr:
We might need this in the future, but not anytime soon. Our API is quite small and we need to focus on the core functionality of the app

GeniJaho:
I know, the only reason why I started integrating this was that Anoop Nair needs some docs to start working with the API. I think he might have a hard time reading the PHP code.

xlcrr:
API not needed anytime soon

Issue #259, opened 2021-07-09
Can't download countries data from the World Cup page


GeniJaho:
![image](https://user-images.githubusercontent.com/29065651/125096071-116dda80-e0d5-11eb-9b4d-7edc041940b7.png)

Issue #170, opened 2021-01-24
About-page, left-aligned or right-aligned


FredMark:
See printscreen.

All text is left-aligned, except for this piece of text?
![Schermafdruk 2021-01-24_21 26 18 (irfanview)](https://user-images.githubusercontent.com/2518047/105642623-1aeae980-5e8b-11eb-827b-643f27cdaf4d.jpg)

GeniJaho:
Is this better?

![image](https://user-images.githubusercontent.com/29065651/125446858-767024ea-d415-491a-b3c5-a176fc7eb598.png)

xlcrr:
This was a feature, not a bug

Issue #148, opened 2021-01-19
The button 'Tag Litter' is not disabled while uploading pictures, if you press it, you miss photos


FredMark:
If you select multiple pictures to upload,
the upload starts,
at the bottom, there is a button 'Tag Litter' that is enabled at that moment.

I pushed it, thinking that you could already start tagging while the upload process was going on in the background.
However, the pictures that you selected to upload are not uploaded, so in the end you miss pictures without knowing it.

Solution:
- disable the button while uploading, enable it (or show it) when the uploads are ready.

Issue #126, opened 2020-12-29
Options under 'More' not visible when the map is shown


FredMark:
See printscreen.
Normally if you go to the menu in the top, and select "More", you get all the options like 'Settings' etc.

However, when the map is shown after you have done this:
(first go to World Cup > choose a country > choose a province > choose a city, so the map is shown.)
If you try to open the menu under "More", the options seem to be 'under' the map, so you can not see them.
![Schermafdruk 2020-12-29_20 15 27 (irfanview)](https://user-images.githubusercontent.com/2518047/103309967-ccc30180-4a16-11eb-97d8-f163277f0521.jpg)

PR #263, opened 2021-07-13
Bugfixes

GeniJaho:
Issue #259 
The issue might be caused by a missing "/" in the request URL.
Through Postman I can confirm that the controller works and the download job is dispatched. However, since I have no initial data and I can't see anything on the World Cup page, I can't say for sure that the Download button works correctly after this fix.

I'm making this a Draft PR, and hopefully, I'll be able to test it after @Xamyrz pushes his changes for the initial setup data that he's working on.

Issue #80 
I updated the post URL, and fixed the backend delete functionality for local/testing environments. I can't test if it works on production. If S3 permissions are tricky and need time to be implemented correctly, we can allow deleting images in local environments, but show the old message "Sorry, this is not working yet" on production.
Also took care to decrement the user's XP and total_images.
Fixed a division-by-zero bug in the user's profile page, which happens when there are no uploaded images, on initial setups.

Issue #170 
Just switched that text from right-aligned to left-aligned. See the issue for the way it looks now.

Issue #148 
Hide the Tag Litter button whenever an upload starts (a single file or multiple), and show it again when uploading finishes (successful or not)

Issue #126 
Simply setting a `z-index: 0` to the map allows other elements to show above it. The map looks the same as before, the z-index doesn't change anything.


GeniJaho:
@xlcrr This PR included a change for that Right-to-Left text on the About page (#170 ), for which you said that it was a feature, not a bug. It's now merged on master and pushed to production. Should I make another commit to revert it to the way it was?

GeniJaho:
I tested the rest of the issues on the live site, I can confirm that everything works fine.
We can download data for countries, states, and cities.

Issue #81, opened 2020-12-10
sitemap


IlanLuci:
It is probably worth adding a sitemap for SEO. I did a quick google search and found a laravel plugin that will automatically crawl the site and generate it. https://github.com/spatie/laravel-sitemap

PR #264, opened 2021-07-13
Added Sitemap Generator command (#81)

GeniJaho:
Using this package https://github.com/spatie/laravel-sitemap, we add the main URLs of the app like the about, world page, etc.
Also, we iterate through all the countries, states, and cities that are manually verified and shown to the users by clicking their links and add them to the sitemap too. I've used the same logic that's applied on the MapController (getCountries, getStates, getCities methods).

To generate the sitemap the first time you need to call `art sitemap:generate`.

To schedule it to run from time to time you need to set up a cron job to call the console Kernel, if not already set up. I've configured the scheduler to run the sitemap generator every day.

I've also run `composer update`, otherwise the package won't install due to old dependencies.


xlcrr (approved):

xlcrr:
We need to upgrade to php 7.4 to use this

PR #269, opened 2021-07-15
Refactor/add locations create service

GeniJaho:
- Removed the methods to extract countries, states, cities from their respective models, to keep the models clean
- Put them into `App\Services\LocationService` 
- I can confirm the bugs with detecting locations came from those `firstOrCreate()` methods, on my implementation it works fine, so I removed the try-catch blocks.
- I've updated the test cases we had, and I used them to confirm that everything still works.
- I've made respective changes to ApiPhotosController, but haven't tested if they work the same.


app/Services/LocationService.php
```diff
@@ -0,0 +1,150 @@
...
+
+        if (!$countryCode) {
+            return Country::where('country', 'error_country')->first();
+        }
+
+        $country = Country::select('id', 'country', 'shortcode')
+            ->firstOrCreate([
+                'shortcode' => $countryCode,
+                'country' => $country
+            ]);
```
xlcrr, line 35: We only want to check for countryCode here

xlcrr (approved):

app/Services/LocationService.php
GeniJaho: True, I forgot about that. In that case, we might want to change the method entirely and not use first or create. Because if we only use the country code, the new country will have an empty full name.

PR #270, opened 2021-07-15
Bugfix - Store tags when switching photos (#67) (#69)

### LINKED ISSUE #67: Not all tags are remembered when you switch to next picture
It is possible to switch to the next picture (very nice!). However, only 1 tag that i gave in the previous photo is still available on the screen. Would be nice to have all the tags.

### LINKED ISSUE #69: Tags that were selected are gone if you skip to previous picture and skip to the next picture
if you are giving tags to a picture, and decide to skip to the previous picture to look at something and then skip back to the current picture, the tags that you already gave are gone.

GeniJaho:
Intended to fix #67 and fix #69, most important change in this PR is modifying the way the tags are kept in store. Instead of the old `{category: { tag: value }}`, we now use `{ photoId: { category: { tag: value, ... }, ... }, ... }`.
This way, we store the tags for every photo.

Since the structure of the `state.tags` has changed, I've checked that these buttons/features still work: Delete photo, Clear Recent Tags, Next/Prev image, Pagination links, clicking on tags to remove them.

Also, I've checked on the profile page, where you can tag multiple images, that feature still works.

On this URL /admin/photos, admin users can verify/delete tagged/untagged images. I've tested that those features still work, except the Verify & Delete button, which isn't implemented yet on the master branch. I noticed we don't update (decrement) the user's XP and total_images when Admin is deleting the photo, is that how it should be?

I've deleted `photo_id` from the coastal table, it causes errors on new setups, and is not needed for category tables.

Let me know if I've forgotten anything. This PR is a good candidate to be tested on a staging server IMO, it contains a lot of changes.


GeniJaho:
This PR can be merged into staging again, after fixing conflicts with the latest merges.

PR #277, opened 2021-07-19
Geni jaho bugfix/UI improvements

xlcrr:


xlcrr:
3 things left to do here @GeniJaho 

1. When the user clicks a tag from global search, it should add the tag.
2. Clear the input
3. Reset the search

GeniJaho:
> 3 things left to do here @GeniJaho
> 
> 1. When the user clicks a tag from global search, it should add the tag.
> 2. Clear the input
> 3. Reset the search

So it should act identically to the Add Tag button?

xlcrr:
Yes, as soon as the click the suggestion, add the tag

GeniJaho:
I added the selected tag when a user selects an option from the global search.
Also, when the page first loads, now the search is on focus, and you can start typing directly.

PR #282, opened 2021-07-23
Dummy data

Xamyrz:
creation of 1500 dummy photos in the area of Cork


xlcrr (changes requested):
added review

app/Http/Controllers/MapController.php
```diff
@@ -121,7 +121,7 @@ public function getCountries ()
         }
 
         /** GLOBAL LITTER MAPPERS */
-	    $users = User::where('xp', '>', 9000)
+        $users = User::where('xp', '>', -1)
```
GeniJaho, line 124: We can just remove the xp condition entirely, given the number of users is low, this query is really fast, so no need to limit by xp. So `User::orderBy('xp', 'desc')->...` should be ok.

app/Models/Location/City.php
```diff
@@ -56,27 +56,29 @@ class City extends Location
      */
     public function getTotalLitterRedisAttribute ()
     {
-        return Redis::hexists("city:$this->id", "total_litter")
+        return (Redis::hexists("city:$this->id", "total_litter"))
             ? (int)Redis::hget("city:$this->id", "total_litter")
-            : 0;
+            : 1;
```
GeniJaho, line 61: We should not modify this, if there's no litter for that key, then simply return 0. To avoid that nasty division by zero error we should make checks where we use this property, for example, MapController, line 68.

app/Models/Location/City.php
```diff
@@ -56,27 +56,29 @@ class City extends Location
...
     /**
      * Return the total_photos value from redis
      */
     public function getTotalPhotosRedisAttribute ()
     {
-        return Redis::hexists("city:$this->id", "total_photos")
+        return (Redis::hexists("city:$this->id", "total_photos"))
             ? (int)Redis::hget("city:$this->id", "total_photos")
-            : 0;
+            : 1;
```
GeniJaho, line 71: Same here, need to revert this.

app/Models/Location/City.php
```diff
@@ -56,27 +56,29 @@ class City extends Location
...
     }
 
     /**
      * Return the total number of people who uploaded a photo from redis
      */
     public function getTotalContributorsRedisAttribute ()
     {
-        return Redis::scard("city:$this->id:user_ids");
+        return (Redis::scard("city:$this->id:user_ids") > 0)
```
GeniJaho, line 79: Revert this too, add checks where it's used.

app/Models/Location/Country.php
```diff
@@ -49,7 +49,7 @@ public function getTotalLitterRedisAttribute ()
     {
         return Redis::hexists("country:$this->id", "total_litter")
             ? (int)Redis::hget("country:$this->id", "total_litter")
-            : 0;
+            : 1;
```
GeniJaho, line 52: Revert this too, add checks where it's used. I think this can be simplified and we can remove that `hexists`, but leave it for now and just revert 1 to 0.

app/Models/Location/Country.php
```diff
@@ -59,15 +59,17 @@ public function getTotalPhotosRedisAttribute ()
     {
         return Redis::hexists("country:$this->id", "total_photos")
             ? (int)Redis::hget("country:$this->id", "total_photos")
-            : 0;
+            : 1;
```
GeniJaho, line 62: Revert this too.

app/Models/Location/Country.php
```diff
@@ -59,15 +59,17 @@ public function getTotalPhotosRedisAttribute ()
...
     }
 
     /**
      * Return the total number of people who uploaded a photo from redis
      */
     public function getTotalContributorsRedisAttribute ()
     {
-        return Redis::scard("country:$this->id:user_ids");
+        return (Redis::scard("country:$this->id:user_ids") > 0)
```
GeniJaho, line 70: Revert this too.

app/Models/Location/State.php
```diff
@@ -39,7 +39,7 @@ public function getTotalLitterRedisAttribute ()
     {
         return Redis::hexists("state:$this->id", "total_litter")
             ? (int)Redis::hget("state:$this->id", "total_litter")
-            : 0;
+            : 1;
```
GeniJaho, line 42: Revert this too.

app/Models/Location/State.php
```diff
@@ -49,15 +49,17 @@ public function getTotalPhotosRedisAttribute ()
     {
         return Redis::hexists("state:$this->id", "total_photos")
             ? (int)Redis::hget("state:$this->id", "total_photos")
-            : 0;
+            : 1;
```
GeniJaho, line 52: Revert this too.

app/Models/Location/State.php
```diff
@@ -49,15 +49,17 @@ public function getTotalPhotosRedisAttribute ()
...
     }
 
     /**
      * Return the total number of people who uploaded a photo from redis
      */
     public function getTotalContributorsRedisAttribute ()
     {
-        return Redis::scard("state:$this->id:user_ids");
+        return (Redis::scard("state:$this->id:user_ids") > 0)
```
GeniJaho, line 60: Revert this too.

GeniJaho (approved):

xlcrr (approved):

Issue #140, opened 2021-01-17
Not possible to download 'all' data


FredMark:
OLM stands for Open Data and indeed it is possible to download all data without obstructions like it should.
However, the max you can download (at a time) is a country.
We should think about some way to download all data, of the entire world.
(If somebody wants 'all data', he has to click on every country to download it,
that is not really facilitating people who want to do research with the data.

xlcrr:
I like the idea of being able to download all global data, but it is an expensive operation
I think we could run this monthly, and make 1 global file available at the start of each month for all months previous

PR #285, opened 2021-07-30
Bugfix/profile and excel data

### LINKED ISSUE #59: Number of XP's received for uploading and tagging pictures not correct
I have tagged and uploaded 1 picture.
I go to the Worldcup, i now have 43999 xp's.
This is 2 more than before. So apparantly 2 xp's for an uploaded and tagged picture?

I upload and tag 3 more pictures,
I go to the Woldcup, i now have 44002 xp's. So uploading and tagging 3 pictures only gives me 1 xp per photo?

I upload and tag 3 extra pictures. In the worldcup i see that i now have 44.009 xp's. So +7 xp's for 3 pictures?

I don't see why this is so 'random', it should be clear to the user what the 'rules' and 'awards' are.
How many xp's do you get for uploading a picture,
How many xp's do you get for tagging, and how many per tag etc.

### LINKED ISSUE #66: Excel does not show if the litter is picked or not
I upload and tag litter,
i have the switch in the bottom right set to 'on' (the litter is picked), and submit a photo and then download the data, you can not find an indication to whether the litter was picked or not, so this data misses in the downloads?

In general:
See different screens, it is not clear what the current setting is 'i picked it up' or 'it is still there'. Some texts are in red as if you are being warned that the setting is wrong, but it is not clear if that is the current setting or if you need to select that one to get that setting active.

### LINKED ISSUE #212: number of uploaded photos and xp not correct at specific points
I have been tagging a lot of clusters of items. (all the same kind of items)
At the start i log all the numbers (like photos uploaded already, number of xp etc,
Then i upload a bunch of photos, wait until all counters are updated and write down the numbers again.
Then i tag all items (for instance a batch with all 2 tags (redbull+energydrink) or 'marlboro+cigarettepack'), wait until all counters are updated and write down the numbers again.

Something is going wrong here:
if i upload 100 photos, the next thing happens (100 is an example, below i will write down some real numbers)
- tag-page says that i have to tag 100 photos (that is Correct) and shows 1, 2, 3, .. 100 in the bottom of the screen --> correct
- Profilepage, number of photos NEVER +100, always less (but not always exactly the same number less)
- Profilepage, number of xp NEVER +100, always less, (and always the same number as the counter of number of photos)
- Worldcuppage, shown number NEVER + 100, always less (and again, always the same number as the counter of aboven.
For instance what i have been uploading and what the numbers where:
uploaded, shown in tagging screen, numberofphotosadded, numberofxpadded:
56, 56, 53, 53
53, 53, 52, 52
37, 37, 35, 35
32, 32, 29, 29

However, when you go to the tagging-screen, there is also a line that says 'total images uploaded'.
If you look at that number and compare it with the counter that was shown before uploading,
this seems to show the correct number: +53, +37 etc.
(At this moment, the tagging screen says 'total images uploaded: 21916, however the profilepage says 21788, so it looks like the profilepage is right, but the other are missing some pictures every time?

### LINKED ISSUE #242: Profilepage not updated after uploading pictures, unless you refresh the page
I have uploaded 28 pictures via the upload page and visit the More>Profile-page after that action.
The figures on that page are 'not correct', they show the figures that were there before uploading the pictures.

As a user, i would expect these numbers to be correct after clicking on that page.
The workaround is to press the 'refresh'-icon, in that case the page is refreshed and now shows the right counters.
(solution: do the refresh automatically if you select the profile-page?)
Schermafdruk 2021-06-12_11 08 40 (irfanview)
Schermafdruk 2021-06-12_11 13 35 (irfanview)

GeniJaho:
Added Picked Up column to CSV exports (Fix #66)
Instead of leaving it as 'remaining', I called it 'Picked Up', and the values are 'Yes' or 'No'.

Fixed wrong XP when uploading multiple photos (Fix #59) (Fix #212)
Basically what was happening was that when a user uploaded many photos, the individual requests could use outdated values for XP and total_images. Using Eloquent's `refresh()` didn't work either, so I used mass assignment and confirmed that it worked correctly.
Also, now the number of total uploaded photos is correct on the Tagging page, it's the same as the number in the Profile page.

Fixed #242 
Profile page gets updated every time you open it, the Radar chart on the left too. No need to refresh the page.

Installed a package called Laravel Debugbar, it's used in almost every Laravel app. It shows a little bar at the bottom of the site, where you can track almost everything, logs, messages, errors, queries, models, etc. It is hidden by default on production.

What is still unfixed is the percentage of tags on the Profile page. When comparing the user's tags to the total number of tags, we don't include brands on the total number of tags, that's why that number is not correct.


xlcrr:
We can trial this global download for a while but the UI needs some work

app/Exports/CreateCSVExport.php
```diff
@@ -660,14 +660,22 @@ public function query ()
...
                 return Photo::with(['smoking', 'food', 'coffee', 'alcohol', 'softdrinks', 'other', 'sanitary', 'brands', 'dumping', 'industrial', 'art'])
                     ->where([
                         'country_id' => $this->location_id,
                         'verified' => 2
                     ]);
             }
+
+            else
+            {
+                return Photo::with(['smoking', 'food', 'coffee', 'alcohol', 'softdrinks', 'other', 'sanitary', 'brands', 'dumping', 'industrial', 'art'])
```
xlcrr, line 674: Does this return all the photos? If we need this we should cluster it. I'm thinking this job is too big...

app/Http/Controllers/ApiPhotosController.php
```diff
@@ -175,6 +176,14 @@ public function store (Request $request) :array
             \Log::info(['ApiPhotosController@store', $e->getMessage()]);
         }
 
+        // Since a user can upload multiple photos at once,
+        // we might get old values for xp, so we update the values directly
+        // without retrieving them
+        $user->update([
+            'xp' => DB::raw('xp + 1'),
+            'total_images' => DB::raw('total_images + 1')
+        ]);
```
xlcrr, line 185: We will migrate this to Redis soon which should fix this issue

app/Http/Controllers/DownloadControllerNew.php
```diff
@@ -64,24 +67,44 @@ public function index (Request $request)
...
         }
     }
+
+    /**
+     * @param string $type
+     * @param int $locationId
+     * @param string $path
+     * @param string $email
+     */
+    protected function export(string $type, int $locationId, string $path, string $email): void
```
xlcrr, line 94: What are we exporting here? Any location? Global? Comment needs a title

app/Http/Controllers/User/ProfileController.php
```diff
@@ -142,12 +139,12 @@ public function index ()
         // Todo - Store this metadata in Redis
         $totalPhotosAllUsers = Photo::count();
         // Todo - Store this metadata in Redis
-        $totalLitterAllUsers = Photo::sum('total_litter');
+        $totalTagsAllUsers = Photo::sum('total_litter'); // this doesn't include brands
```
xlcrr, line 142: We also need to store this on Redis

app/Models/Photo.php
```diff
@@ -278,75 +279,75 @@ public function city ()
      */
     public function smoking ()
     {
-    	return $this->hasOne('App\Models\Litter\Categories\Smoking', 'id', 'smoking_id');
+    	return $this->belongsTo('App\Models\Litter\Categories\Smoking', 'smoking_id', 'id');
```
xlcrr, line 282: A photo has one category : smoking ?

resources/js/store/modules/user/actions.js
```diff
@@ -146,8 +142,6 @@ export const actions = {
             }
         })
         .then(response => {
-            console.log('get_users_profile_map_data', response);
```
xlcrr, line 149: Please put back these comments, the lowercase action name is very useful when debugging and checking incoming data locally

resources/js/views/Locations/SortLocations.vue
```diff
@@ -1,5 +1,12 @@
 <template>
     <section class="locations-main" :class="container">
+
+        <div class="columns py-4 px-4">
+            <div class="column is-half is-offset-3">
+                <Download type="worldwide"/>
```
xlcrr, line 6: Small change, but call this "global" pls

resources/js/views/Locations/SortLocations.vue
```diff
@@ -194,7 +201,7 @@ export default {
 
 	.locations-main {
 		background-color: #23d160;
-		min-height: 100%;background-color: #23d160; 
+		min-height: 100%;background-color: #23d160;
```
xlcrr, line 204: background-color is duplicated here

resources/js/views/general/Tag.vue
```diff
@@ -87,7 +87,7 @@
                                     {{ $t('tags.to-tag') }}: {{ remaining }}
                                 </li>
                                 <li class="list-group-item">
-                                    {{ $t('tags.total-uploaded') }}: {{ user.photos_count }}
+                                    {{ $t('tags.total-uploaded') }}: {{ user.total_images }}
```
xlcrr, line 90: We should be getting this data from Redis

xlcrr (changes requested):

app/Http/Controllers/ApiPhotosController.php
GeniJaho: Upgrading this to Redis might take some time I think, even if we prioritize it. I'd like to do the migration using tests, I feel it's going to introduce a lot of new bugs otherwise. For that reason, this update() operation fixes the issue for now.

app/Exports/CreateCSVExport.php
GeniJaho: Yes, this returns all the photos where verified = 2. What do you mean by cluster it? Indeed, the job is big, can we try it on the production server, without allowing the users to interact with it? We can never know for sure unless we try it.

app/Http/Controllers/DownloadControllerNew.php
GeniJaho: I'm adding a comment on this.

app/Http/Controllers/User/ProfileController.php
GeniJaho: True, I noticed the Todo there, but I wanted to keep this PR simple and fix the issues with the current codebase. Migrating to Redis would introduce some complexity which I'm not sure we can handle without its own PR.

resources/js/store/modules/user/actions.js
GeniJaho: Ok, I'm bringing them back.

resources/js/views/Locations/SortLocations.vue
GeniJaho: Ok, this is actually not used, it's there only because the Download component requires a type.

resources/js/views/Locations/SortLocations.vue
GeniJaho: It's now fixed from merging staging here.

resources/js/views/general/Tag.vue
GeniJaho: Yes, but the user's total_images is used elsewhere on the site too. We can get it from Redis when we migrate the other parts too. Besides, we can still use total_images as a user property, and change the implementation, or data source, on the user model on the backend, without changing anything here.

GeniJaho:
> I like the idea of being able to download all global data, but it is an expensive operation
> I think we could run this monthly, and make 1 global file available at the start of each month for all months previous
> The UI needs some improvement before merging

True, it is an expensive operation, but I think the server can handle it. Can we give it a shot without allowing the users to do it?
I updated the PR to store the global CSV files every month, not every day. If it works as-is on production, of course. In any case, I can find a more efficient solution if needed.
If we go forward with this, and I think we should since Fred made a good point on issue #140 , what UI improvements are on your mind? We can create a new component from this, although I'd hate to duplicate the logic we currently use.

app/Models/Photo.php
GeniJaho: Using `hasOne` doesn't work well when retrieving things like `$photo->smoking`, I noticed that when I was refactoring into actions, those lines with `$photo->fresh()->$category()->create()`. The relationship type should reflect the way foreign keys are placed between tables, and, in our case, since a photo has a smoking_id, the photo belongs to smoking. HasOne would make sense if smoking had a photo_id.

Anyway, the database design could be improved in my opinion. The right way to model the relationship photo - category would be a [Many to Many](https://laravel.com/docs/8.x/eloquent-relationships#many-to-many). This is true because a photo can be related to many categories, and a category would be applied to many photos.
This would result in a new table, called photos_categories, which would connect the two types of models. In turn, we would have a separate table for `tags -> (category_id, tag_name)`. This way we'd have a more normalized database, and it would be easier to make changes, and if I'm not wrong is the reason why we're migrating to Redis, right?

app/Exports/CreateCSVExport.php
xlcrr: I meant chunking. Instead of getting all data, we can get it in chunks of 5,000 photos for example

GeniJaho: I'm removing this feature from this PR entirely, let's think about it sometime later.

app/Models/Photo.php
xlcrr: The belongTo now makes sense to me. Nice find. I like the sound of the new photos_categories and tags table! I think we will keep redis mostly for metadata like total_photos, leaderboards, and keep the photos + tags data on mysql

PR #304, opened 2021-08-05
Create laravel.yml Github Action

GeniJaho:


xlcrr:
What does this actually do ?

GeniJaho:
Every time we push or make a PR to master or staging, Github will execute our tests suite, and will add a green checkmark if they pass. This way everyone knows if they've broken something, even if they haven't run the tests locally.

PR #306, opened 2021-08-06
added previous_tags setting

jshepin:
this PR creates a setting for users to allow previous tags used in an image to appear on the current one.


resources/js/components/Litter/AddTags.vue
```diff
@@ -197,6 +209,11 @@ export default {
         };
     },
     computed: {
+        
+        //whether the user wants to show previous tags. return type: boolean
+        show_previous_tags(){
+            return this.user.previous_tags == 1;
```
xlcrr, line 215: When the value of `previous_tags` is a boolean or number, we can return that if it is `1`. It helps to wrap these statements in brackets too to show its computing something. Can you update the comment block to match the others too?

```
/**
 * Return true to show previous tags
 */
show_previous_tags () 
{
     return (this.user.previous_tags);
}
```

resources/js/components/Litter/AddTags.vue
```diff
@@ -543,6 +576,7 @@ export default {
          */
         async submit ()
         {
+            // this.previousTags = (this.$store.state.litter.tags[this.id] || {});
```
xlcrr, line 579: Do we need this or not? If its not needed it can be deleted

resources/js/store/modules/user/actions.js
```diff
@@ -232,6 +232,32 @@ export const actions = {
         });
     },
 
+    /**
+     * Save all tag settings on Tagging.vue
+     */
+    async SAVE_TAGGING_SETTINGS (context)
+    {
+        let title = i18n.t('notifications.success');
+        let body  = i18n.t('notifications.tagging-updated');
```
xlcrr, line 241: When the variable does not change, use `const`

resources/js/store/modules/user/mutations.js
```diff
@@ -119,6 +119,14 @@ export const mutations = {
         state.user[payload.column] = payload.v;
     },
 
+    /**
+     * The user wants to change a tagging setting
+     */
+    tagging (state, payload)
+    {
+        state.user[payload.column] = payload.v;
```
xlcrr, line 127: This is a dynamic function that is being repeated. We should either use 1 dynamic function and pass whatever keys we want to update, or else have static functions that only change a specific param

resources/js/views/general/Tag.vue
```diff
@@ -168,6 +168,11 @@ export default {
     },
     computed: {
 
+        //whether the user wants to show previous tags. return type: boolean
+        show_previous_tags(){
+            return this.user.show_previous_tags == 1;
+        },
```
xlcrr, line 174: Function can be updated similar to the other. Do we need to duplicate them? We could pass the value from one component to the other and accept is as a prop

resources/js/views/general/Tag.vue
```diff
@@ -231,7 +236,8 @@ export default {
         user ()
         {
             return this.$store.state.user.user;
-        }
+        },
+
```
xlcrr, line 240: delete `,` and extra line

xlcrr (changes requested):

resources/js/components/Litter/AddTags.vue
jshepin: Yes we do. When you click the submit button, then it stores the tags you used.

resources/js/store/modules/user/actions.js
jshepin: good catch, I fixed them all in `actions.js`

resources/js/views/general/Tag.vue
jshepin: I got rid of the duplicate. I forgot to get rid of it in a previous commit.

resources/js/components/Litter/AddTags.vue
xlcrr: Is it meant to be commented out?

app/Http/Controllers/UsersController.php
```diff
@@ -344,6 +344,19 @@ public function togglePrivacy (Request $request)
...
+     * Toggle a Users tagging settings
+     */
+    public function toggleTagging (Request $request)
+    {
+        $user = Auth::user();
+
+        /* Whether to save the tags from a previous image */
+        $user->previous_tags = $request->previous_tags;
+
+        $user->save();
```
xlcrr, line 357: We should also return a response and handle it 

```php
return [
    'success' => true
];
```

```javascript

.then(response => {
   console.log('my_action', response); 
   
    if (response.data.success)
    {
          // Handle success eg show successful event notification
          
```

resources/js/components/Litter/AddTags.vue
```diff
@@ -72,11 +71,26 @@
             </div>
 
             <br><br>
+         
+           <div v-if="show_previous_tags && Object.keys(previousTags).length > 0 && show_previous_tags">
+            <h1>{{ $t('tags.previous-tags') }}:</h1>
```
xlcrr, line 76: Has this been deleted? I'm still seeing it in the PR

xlcrr (changes requested):

resources/js/components/Litter/AddTags.vue
jshepin: I got rid of it. My bad.

app/Http/Controllers/UsersController.php
jshepin: good catch

PR #312, opened 2021-08-08
fix bbox tagging 

jshepin:
addresses both issues addressed in this Trello ticket https://trello.com/c/cfYREOHA/266-fix-broken-tagging-on-bbox


xlcrr:
We want to keep label false by default

jshepin:
Ok I'll make the change back. I made it true by default because I found it difficult to move my eyes back and forth on the screen and it makes sense to me to put the label on the image. Just my opinion though.

jshepin:
image
could we just make the "toggle label" button stateful so you don't have to press it every time?

PR #313, opened 2021-08-08
Feature - Fix & improve map images

GeniJaho:
I've increased the size of the images, both art and regular types, on all maps (Global map, Profile, Teams, City Maps). All the popups have the same styling and format, in contrast to what we had before that we listed tags above the image on Profile, Teams, and City maps. I think it makes for a better experience to have the same layout on all maps.

I also increased the font size of the popup text a little bit, it really is hard to read and very small.


resources/js/views/global/Supercluster.vue
```diff
@@ -184,177 +183,23 @@ function onEachArtFeature (feature, layer)
...
+                )
             )
             .openOn(map);
     });
 }
 
-/**
- * The user dragged or zoomed the map, or changed a category
- */
-async function update ()
```
xlcrr, line 213: We want to separate some Vue and Map state. This update was called by change map. Does it still work moved back into vue methods?

xlcrr (approved):

resources/js/views/global/Supercluster.vue
GeniJaho: I actually wanted to move some logic into Vue instance methods to possibly use Vue reactivity to optimize things a bit. However, I've only almost copy-pasted the method without changing anything, and I've tested that it works just as it did before.

I think that the code for the maps still needs a lot of refactoring, and moving code from functions into objects is a good step in the right direction. Since the code for the Global map is different from the one for Profile, Teams, and Cities I didn't change it, just moved a part of it into Vue. What do you think?

xlcrr: Using more vue would be great but I found it tricky to combine leaflet and vue. If we have to keep them separate, we could create another mapFunctionHelper

PR #314, opened 2021-08-14
New Setting to make Profile Public, new UserSettings table

xlcrr:


GeniJaho:
I propose we use a package to handle user settings. This is both for new settings added and migrating the old ones. Having them as database columns makes it pretty hard to update them.
I propose we use this package https://github.com/DarkGhostHunter/Laraconfig which I think is more suitable for our use-case; another option is https://github.com/glorand/laravel-model-settings.

PR #327, opened 2021-08-24
Bugfix - Teams

GeniJaho:


app/Providers/EventServiceProvider.php
```diff
@@ -50,14 +53,10 @@ class EventServiceProvider extends ServiceProvider
             // 'App\Listeners\GenerateLitterCoin',
             // 'App\Listeners\UpdateLeaderboardsAdmin', happens on AddTagsTrait
             'App\Listeners\AddTags\CompileResultsString',
-            // todo - only call this listener if the user has active_team
-            'App\Listeners\AddTags\IncrementUsersActiveTeam',
+            IncreasePhotoTeamTotalLitter::class,
             'App\Listeners\User\UpdateUserTimeSeries',
             'App\Listeners\User\UpdateUserCategories'
         ],
-        TagsDeletedByAdmin::class => [
```
GeniJaho, line 58: @xlcrr We actually don't need to decrement the Redis location values when an admin deletes tags from a photo. That is because if an unverified user submits tags for a photo, we don't call the TagsVerifiedByAdmin event, and don't run the IncrementLocation listener. For that reason, we shouldn't execute DecrementLocation when these tags are deleted. Moreover, admins can only delete tags from unverified photos.

PR #332, opened 2021-08-26
Feature - Ability to leave a team, inactivate all teams

GeniJaho:
The backend functionality is fully tested. These conditions are in place:
- Any user can leave a team
- A user can only leave a team they are part of
- If the user leaves their active team, the next team they are part of will become their new active team. Otherwise, if they leave their only team, then `$user->active_team = null`
- A user can not leave a team if they are the only member, this is to prevent teams without a leader, or members
- If the leader of a team leaves, the next member of that team will become the new leader
- When a user leaves a team, their contributions to that team are reset. More precisely we delete the relation on `team_user`, which contains `team_user.total_photos` and `team_user.total_litter`. No other team info is changed. For example, if a user uploads 3 photos and submits 5 tags for a team; when they leave, the team will still have these photos and tags, but only the user contribution will be removed (total_litter and total_photos). So if this user re-joins the team, these values will be 0.

- When a user creates a team, their active team now becomes this new team. If they go to My Teams they will see the new team as selected.

Also, a user can now inactivate their current active team. This means although they will be part of some teams, their progress will not be recorded in any of them.


app/Http/Controllers/Teams/LeaveTeamController.php
```diff
@@ -0,0 +1,101 @@
...
+use Illuminate\Support\Facades\DB;
+
+class LeaveTeamController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('auth');
+    }
+
+    public function __invoke(Request $request)
```
xlcrr, line 18: Comment missing. What does this do?

app/Http/Controllers/Teams/LeaveTeamController.php
```diff
@@ -0,0 +1,101 @@
...
+            'teamId' => 'required|exists:teams,id'
+        ]);
+
+        /** @var User $user */
+        $user = auth()->user();
+        /** @var Team $team */
+        $team = Team::find($request->teamId);
+
+        if (!$user->teams()->whereTeamId($request->teamId)->exists()) {
+            abort(403, 'You are not part of this team!');
```
xlcrr, line 30: I would prefer to `return ['success' => false, 'msg' => '']` and handle it with a popup / toast notification

resources/js/store/modules/teams/actions.js
```diff
@@ -27,13 +27,59 @@ export const actions = {
...
 
+    /**
+     * Leave the team
+     */
+    async LEAVE_TEAM (context, payload)
+    {
+        await axios.post('/teams/leave', {
+            teamId: payload
+        })
+        .then(response => {
```
xlcrr, line 46: after every request, please add a log with the name of the action in lowercase. I find this makes debugging easy to find out what requests are being made and what actions are being called

resources/js/store/modules/teams/actions.js
```diff
@@ -27,13 +27,59 @@ export const actions = {
...
+            if (response.data.success) {
+                Vue.$vToastify.success({
+                    title: i18n.t('notifications.success'),
```
xlcrr, line 49: There is a bug with i18n. During long wait times, the translation doesn't always happen. We need to declare the translations before doing `await`

xlcrr (changes requested):
1-2 small changes needed

app/Http/Controllers/Teams/LeaveTeamController.php
GeniJaho: These abort operations are meant to handle cases where users hackishly try to make requests. So they should never appear from users doing normal things. That's why we exit with an exception (aborting).

app/Http/Controllers/Teams/LeaveTeamController.php
GeniJaho: You mean the controller as a whole or the invoke method? If you mean the invoke method, no need to comment it, they're pretty common, see https://laravel.com/docs/8.x/controllers#single-action-controllers

GeniJaho:
I implemented the fixes for i18n and console logs after requests.

PR #334, opened 2021-08-27
Feature - Update team name and identifier

GeniJaho:
Added a section in the Team Settings screen where a user can update the names and identifiers of all the teams where they are the leader. All the new strings added have their translations too. It looks like this (image is zoomed out to capture the whole screen):

image

The back-end functionality is fully tested. It has the same validation rules as when you create a team.
If the user is not a leader in any team, then this whole section will not show at all.
Also, when a user is not part of any team, neither section will be shown. Instead, I've put the same text as in My Teams screen:

image


app/Http/Controllers/Teams/LeaveTeamController.php
```diff
@@ -0,0 +1,101 @@
...
+use Illuminate\Support\Facades\DB;
+
+class LeaveTeamController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('auth');
+    }
+
+    public function __invoke(Request $request)
```
xlcrr, line 18: comment missing

xlcrr (changes requested):
We need to move the translations outside of the requests due to a bug with i18n

PR #338, opened 2021-08-31
Bugfixes - Teams

### LINKED ISSUE #107: Set of issues when showing the tags on the picture
Once you have tagged a picture, you can find the picture on the map and it will show the tags that are registered on that picture.
Well, that is, móst of the tags are shown, but not all.

Below the issues that i found while giving all tags that were possible on Web (there are some more tags possible when you use the app for tagging, so these have not been tested):

- Food: nr 9 is missing (sauce packet, this one did show up in the excel)
- Industrial: nr 2 & 6 missing ('plastic' and 'other'. They were also missing in the excel)
- Other: 2, 3, 4, 5, 6, 10 & 14 missing (2, 3, 4, 5, 6, 10 did show up in excel. 14 did also not show up in excel. 7, 8 & 9 did not show up in excel)
- Sanitary: nr 7 (ear swabs) missing. It did show up in excel.
- Softdrink: nr 19, 20, 21 & 22 missing. They did show up in excel. Nr 1 did not show up in excel.
- Smoking: OK
- Brands: OK
Schermafdruk 2020-12-20_23 07 25 (irfanview)
Schermafdruk 2020-12-20_23 08 16 (irfanview)
Schermafdruk 2020-12-20_23 09 58 (irfanview)
Schermafdruk 2020-12-20_23 10 42 (irfanview)
Schermafdruk 2020-12-20_23 11 29 (irfanview)

### LINKED ISSUE #120: Team-Screen too small, could be bigger
See printscreen,

i wanted to see more on the map, so i zoomed out.
However, it looks like there is a designated place for the map while it could be much bigger.
Schermafdruk 2020-12-26_15 51 29 (irfanview)

GeniJaho:
https://trello.com/c/sCFnYdoP
https://trello.com/c/9AMUmnbz
Fix #120 
Instead of making the map larger, I simply added a button to toggle fullscreen mode, similar to profile page maps.

https://trello.com/c/QUpf5YFk
Fix #107 
Did some refactoring and automated testing for this one:
- For all categories, I pulled the types() method up to their parent LitterCategory, which is now an abstract class
- It still returns the same data, since we have tests to prove that it is equal to the fillable attribute of each category
- I suggest we remove the fillable property for categories, no need to duplicate the tags listing anymore, we can simply set `protected $guarded = [];`. This can be easily done now.
- Introduced a new method typesForExport() on each category. This method returns an array where keys are the tags and values are the strings used in the CSV export. 
- Since we're using typesForExport() now, we can easily put translated tags instead of hardcoded strings
- I changed the order of tags and categories in the CSV export. Now the Other category is last, next to Art. I also introduced the Dogshit category to the export, it was missing.
-  Now we have dogpoo and dogpooinbag both in Dogshit and Other category. I think this might be the issue with wrong totals on excel files, since we might have dogpoo data in both categories. We can check that after this is merged.
- Fixed and added tests to make sure that all tags on the models are identical to the database columns

Next step in this direction is unifying the tags and categories on the front-end, but that might be a bit more difficult.


app/Models/Litter/Categories/Alcohol.php
```diff
@@ -27,24 +27,21 @@ public function photo () {
...
+            'spiritBottle' => 'spirit_bottle',
+            'wineBottle' => 'wine_bottle',
+            'brokenGlass' => 'broken_glass',
+            'bottleTops' => 'beer_bottle_top',
+            'paperCardAlcoholPackaging' => 'paper_card_alcohol_packaging',
+            'pint' => 'pint_glass',
+            'plasticAlcoholPackaging' => 'plastic_alcohol_packaging',
+            'six_pack_rings' => 'six_pack_ring',
+            'alcohol_plastic_cups' => 'alcohol_plastic_cup',
+            'alcoholOther' => 'alcohol_other',
```
xlcrr, line 44: ideally we should only have 1 key for everything which would make it more dynamic

app/Models/Litter/Categories/Brand.php
```diff
@@ -106,110 +107,88 @@ class Brand extends LitterCategory
...
+            'starbucks' => 'starbucks',
+            'tayto' => 'tayto',
+            'tesco' => 'tesco',
+            'thins' => 'thins',
+            'volvic' => 'volvic',
+            'waitrose' => 'waitrose',
+            'walkers' => 'walkers',
+            'woolworths' => 'woolworths',
+            'wilde_and_greene' => 'wilde_and_greene',
+            'wrigleys' => 'wrigleys',
```
xlcrr, line 191: seems like a lot of duplication

xlcrr (changes requested):
I'm not a fan of this duplicate way of declaring things twice 

Happy to use unguarded, but the types array is used in a few places

app/Models/Litter/Categories/Alcohol.php
GeniJaho: We now have automated tests that prove that the types() array is identical to the db table columns, so there is no risk there at all.
We can totally remove these other keys, but they're not duplicates, they're used in data exports. We can directly use tags instead of these adaptations in exports. Actually, what I think makes more sense, is to completely remove these tag names, and on the CSVExporter use translations instead.
But for the time being we need them. Putting them here makes it easier to test their correctness, and remove the cluttering from csv exporter.

app/Models/Litter/Categories/Brand.php
GeniJaho: It's the only way to automatically test that the columns on the CSV exports are the same as the tags. It's a necessary duplication for now to assert the data correctness. Of course, I intend to refactor this even further by introducing translations and most likely go back to the single types() array, but this is a necessary step in that direction.

And by the way, we still have and use the types() array on every category, it's declaration is simply refactored up to their parent LitterCategory. And it returns the same identical data as before.

app/Models/Litter/Categories/Alcohol.php
xlcrr: Translations would be good, but isn't the types array enough to generate the csv export? This seems like a lot of duplication... it would be cool if we only have to declare each key once

GeniJaho: We can just go back to the old types() method, but would you be ok with the csv column names to be identical to the db column names? After all, you added the names on the csv to be more expressive, right? When we add translations later to the csv we'll remove this duplication entirely, that's for sure. But with these changes we'll be able to write tests for that.

xlcrr: I'm happy for them to be the same as the table column names for now. The problem with adding more code like this is it makes adding new items more difficult. I don't think we will be adding these translations anytime soon. If anyone wants them in their language they will have to open a PR, or see our js translations to match key => value for now

GeniJaho: Ok cool, I'm fixing it then.

GeniJaho: @xlcrr I refactored the types array. I also replaced fillable with the $guarded property, on the LitterCategory abstract class, so all the others inherit it, I wrote a test that it works for all categories.
I also deleted the photos() relationship on some of the models that had it. The proper type of relationship depending on the DB foreign keys should have been Smoking->hasOne(Photo), but even that was not working. It wasn't working in any of them, and it's not used anywhere in code.

app/Models/Litter/Categories/Brand.php
```diff
@@ -120,15 +17,13 @@ public function types ()
             'applegreen',
             'asahi',
             'avoca',
-
```
xlcrr, line 123: Keep the spaces between the first letter please. This was included for a reason- to make adding new ones easier

app/Models/Litter/Categories/Brand.php
```diff
@@ -198,18 +79,15 @@ public function types ()
...
             'thins',
-
             'volvic',
-
             'waitrose',
             'walkers',
             'woolworths',
             'wilde_and_greene',
-            'wrigleys'
+            'wrigleys',
```
xlcrr, line 90: please remove the trailing commas

app/Models/Litter/Categories/Dogshit.php
```diff
@@ -13,20 +13,11 @@ class Dogshit extends LitterCategory
...
-    protected $fillable = [
-        'poo',
-        'poo_in_bag'
-    ];
-
-    /**
-     * Pre-defined litter types available on this class
-     */
-    public function types ()
+    public function types(): array
```
xlcrr, line 16: It would help to have these as static methods. We can then use $types = Photo::types(); anywhere to get the types

app/Models/Litter/Categories/Drugs.php
```diff
@@ -2,35 +2,26 @@
 
 namespace App\Models\Litter\Categories;
 
-use Illuminate\Database\Eloquent\Model;
+use App\Models\Litter\LitterCategory;
 
-class Drugs extends Model
+class Drugs extends LitterCategory
```
xlcrr, line 7: This category can be deleted

app/Models/Litter/Categories/Political.php
```diff
@@ -2,9 +2,12 @@
 
 namespace App\Models\Litter\Categories;
 
-use Illuminate\Database\Eloquent\Model;
+use App\Models\Litter\LitterCategory;
 
-class Political extends Model
+class Political extends LitterCategory
```
xlcrr, line 7: This category can be deleted

resources/js/components/Teams/TeamMap.vue
```diff
@@ -71,6 +79,22 @@ export default {
...
+            this.fullscreen = fullscreen
+        },
+
+        /**
+         *
+         */
+        toggle ()
+        {
+            this.$refs['fullscreen'].toggle() // recommended
+        },
```
xlcrr, line 97: Comments missing or delete the comment blocks

tests/Unit/Models/Litter/LitterCategoryTest.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * @dataProvider getCategories
+     * @param $category
+     */
+    public function test_categories_have_identical_types_as_their_schema($category)
+    {
+        $photo = Photo::factory()->make();
+        $model = $photo->$category()->make([]);
+        $types = $model->types();
```
xlcrr, line 27: We should use a static method to get types, same as we do for categories

xlcrr (changes requested):

app/Models/Litter/Categories/Brand.php
GeniJaho: I generally like to keep these trailing commas in arrays like these. I've seen it in other open-source repositories too. They do it so that when you add a new item, the git diffs show only a new line added, not two lines changed. I can remove the comma, but I think it's better like this. Should I remove them?

resources/js/components/Teams/TeamMap.vue
GeniJaho: Deleting the comments.

app/Models/Litter/Categories/Political.php
GeniJaho: While we don't use these tags anymore, there might be existing photos tagged with Drugs or Political tags. If we're going to delete a category model, shouldn't we delete the DB columns on photos too, or even the drugs and political table?
I think we can push this deletion for a little later. Staging is already way ahead of master.

app/Models/Litter/Categories/Dogshit.php
GeniJaho: We can't do that right away. In order to get the types for a category, we need the category class name. But on a photo model, we only have the categories() method as a static method. The only current way to get to the class name of the categories is by instantiating them, like `$photo->$category->make();`. Otherwise, we'll have to construct a method that returns class names, like this:
```php
 public static function getCategoryClassName($category): string
    {
        $classNames = [
            'smoking' => Smoking::class,
            'art' => Art::class,
            ...
        ];
        return $classNames[$category];
    }
```
So that we can call `::types()` on it. But I don't like this approach at all. It would make the Photo model more complex than it is, we'd be duplicating the categories. I converted all the types() methods on the categories into static methods, but if we try to use them statically it will be a step back in refactoring. I propose we keep them as static methods but don't change the other parts of the code because of that. What do you think?

tests/Unit/Models/Litter/LitterCategoryTest.php
GeniJaho: In this case we need the `$model` to get the `$model->tableName()` further below. I converted the types() on the categories into static methods, but since we have the categories `$model`, we can still call types() from it.

app/Models/Litter/Categories/Brand.php
xlcrr: I don't mind. 1 new line in git diff makes sense

app/Models/Litter/Categories/Political.php
xlcrr: Good point. I am 99% sure they are not used. I will take a look, not important for now

app/Models/Litter/Categories/Art.php
```diff
@@ -8,14 +8,10 @@ class Art extends LitterCategory
 {
     protected $table = 'arts';
 
-    protected $fillable = [
-        'item'
-    ];
-
-    public function types()
+    public static function types(): array
```
xlcrr, line 11: This is a static function

app/Models/Litter/Categories/Brand.php
```diff
@@ -6,110 +6,7 @@
...
-        'woolworths',
-        'wilde_and_greene',
-        'wrigleys'
-    ];
-
-    /**
-     * Pre-defined litter types available on this class
-     */
-    public function types ()
+    public static function types(): array
```
xlcrr, line 9: This is also static

xlcrr (approved):
We are using static again? Sounds good to me

PR #339, opened 2021-09-01
Bugfixes - Reset Password

GeniJaho:
- Fixes layout


resources/js/store/modules/user/actions.js
```diff
@@ -30,6 +31,54 @@ export const actions = {
...
+    {
+        const title = i18n.t('notifications.success');
+
+        await axios.post('/password/email', {
+            email: payload,
+        })
+        .then(response => {
+            console.log('send_password_reset_link', response);
+
+            Vue.$vToastify.success({title, body: response.data.message});
```
xlcrr, line 47: We should prob check for a successful response before showing a notification? I find its good practice to always return `['success' => true]` when a response has completed successfully

xlcrr (approved):

xlcrr:
@GeniJaho can you recompile this please so we can test it on staging

resources/js/store/modules/user/actions.js
GeniJaho: @xlcrr Ok, I'm adding the extra check.
However, if the response was successful then it will return an OK status code. If it wasn't for any reason, the status code should indicate that too. If we are to use 'success' => true and false, we still return a response with the proper status code. In Vue, if it was successful we enter the .then block, else if we got status 422 we enter the .catch block. So the extra check on the .then block is unnecessary.
However, only this and only a few other routes are implemented like this, most of the routes return 200 OK with a status => false, it would be better if they returned a proper status code, and we could omit the success output at all.

xlcrr: You are right about the status code. I didn't have much time to think about that when developing it, and for now failed try/catch statements return 200 success => false in a few places, if they fail. It might be the right thing to do moving forward, I prefer the reading of .success personally

PR #345, opened 2021-09-13
Added data validation when adding tags

GeniJaho:
https://trello.com/c/rnq4vvr8

- Created two new FormRequests, they're different since mobile API and web API are different.
- Unified the checks we do on both endpoints.
- Added tests for the validated fields.

### DIFF
--- a/app/Http/Requests/AddTagsApiRequest.php
+++ b/app/Http/Requests/AddTagsApiRequest.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Http\Requests;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+class AddTagsApiRequest extends FormRequest
+{
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'photo_id' => 'required|exists:photos,id',
+            'litter' => 'required|array'
+        ];
+    }
+}
--- a/app/Http/Requests/AddTagsRequest.php
+++ b/app/Http/Requests/AddTagsRequest.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Http\Requests;
+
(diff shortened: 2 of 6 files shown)


app/Http/Requests/AddTagsApiRequest.php
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'photo_id' => 'required|exists:photos,id',
+            'litter' => 'required|array'
```
xlcrr, line 18: key should be tags. Is it meant to be an array?

app/Http/Requests/AddTagsRequest.php
```diff
@@ -0,0 +1,22 @@
...
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'photo_id' => 'required|exists:photos,id',
+            'tags' => 'required|array',
+            'presence' => 'required|boolean'
```
xlcrr, line 19: This might be undefined

xlcrr (commented):
See comments

app/Http/Requests/AddTagsApiRequest.php
GeniJaho: It would be better to call it tags, like in the web version, but since the app uses the 'litter' key we can't change it.
We could maybe accept both 'litter' and 'tags' for some time, until all the app users have updated their apps to use 'tags'. But I guess we can do that later.
Yes, it's meant to be an array.

app/Http/Requests/AddTagsRequest.php
GeniJaho: If we don't pass the 'presence' parameter the code will attempt to store null on a non-null column and will throw an exception. My argument is that we should also add this parameter on mobile API.

PR #348, opened 2021-09-14
Merge Staging

xlcrr:


app/Http/Controllers/ApiPhotosController.php
```diff
@@ -264,11 +244,13 @@ public function check ()
...
-
-        if ($photos) return ['photos' => $photos];
+        $photos = $user->photos()
+            ->where('verified', 0)
+            ->where('verification', 0)
+            ->select('id', 'filename')
+            ->get();
 
-        return ['photos' => 'none'];
+        return ['photos' => $photos];
```
xlcrr, line 253: Is this still used? Seems unusual to return an array of all photos. Could be large.

GeniJaho: We already return an array of all photos, but apparently, that has never been a problem since we return only one user's photos. This part of code that returns 'none' was never reached actually, the `if` condition was wrong. If there are no photos then the $photos array will be empty.

config/app.php
```diff
@@ -123,6 +123,8 @@
 
     'cipher' => 'AES-256-CBC',
 
+    'root_dir' => env('APP_ROOT_DIR', '/home/vagrant/Code/olm'),
```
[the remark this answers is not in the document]
GeniJaho: We already have checks for prod environment where this is used. It will run the same in every environment with this change. We should refactor this further but I was just getting annoyed by having to change it countless times.

routes/api.php
```diff
@@ -55,17 +52,10 @@
...
-Route::post('/photos/update', 'ApiPhotosController@dynamicUpdate')
-    ->middleware('auth:api');
-
 // Check for any photos uploaded on web
 Route::get('/check-web-photos', 'ApiPhotosController@check')
     ->middleware('auth:api');
 
-Route::post('/add-tags-to-web-img', 'ApiPhotosController@confirm')
-    ->middleware('auth:api');
-
```
xlcrr, line 68: These are definitely the correct routes to delete? I need to double check this on mobile

xlcrr (commented):
2 things to check on mobile

app/Http/Controllers/ApiPhotosController.php
GeniJaho:
```php
        $photos = $user->photos()->where('verification', 0)->select('id', 'filename')->get();

        if ($photos) return ['photos' => $photos];

        return ['photos' => 'none'];
```
This was the original condition. You'll notice `$photos` is a collection instance, which means even if it's empty, `if ($photos)` will always be true. This means we never return the 'none' string, if `$photos` is empty we simply return an array. My refactoring does not change anything, it simply removes the unreached code.

app/Models/Litter/Categories/Alcohol.php
```diff
@@ -44,7 +22,7 @@ public function types ()
             'plasticAlcoholPackaging',
             'six_pack_rings',
             'alcohol_plastic_cups',
-            'alcoholOther'
+            'alcoholOther',
```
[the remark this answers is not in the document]
GeniJaho: You already reviewed the PR for these changes, and we agreed to keep the commas because they make for clean git diff views, so if we add a new item we'll get only a new line added, not two lines changed. This is a good practice used in most open-source projects. I suggest we keep them, are you sure we really need to remove them?

routes/api.php
GeniJaho: I asked Anoop if this route was ever used in the app, and it wasn't. But yeah, better double-check this anyway.

resources/js/store/modules/user/actions.js
```diff
@@ -44,8 +44,6 @@ export const actions = {
         .then(response => {
             console.log('send_password_reset_link', response);
 
-            if (!response.data.success) return;
```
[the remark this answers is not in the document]
GeniJaho: I don't understand this, really. What more do you want to do after the user requests a link, other than showing them that we've emailed their link?
The current implementation returns a redundant 'success' attribute since there are only two possible outputs from the API, either successful or with errors. If we have a successful response, we'll go to the `then` block, and the 'success' parameter is always true in this case, from the backend. That's why this condition is unnecessary, you can't enter it whatever happens. If we have an error, we automatically enter the `error` block below, which handles the error and shows it to the user.

app/Mail/ContactMail.php
```diff
@@ -0,0 +1,48 @@
...
+    public $message;
+
+    /**
+     * Create a new message instance.
+     *
+     * @return void
+     */
+    public function __construct(string $subject, string $message, ?string $name, string $email)
+    {
+        $this->subject = $subject;
```
xlcrr, line 33: there is no subject public var

resources/js/store/modules/user/actions.js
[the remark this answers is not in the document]
xlcrr: Sometimes its usual to return false from a controller. I find the success return method a good strategy to always show the intended success notification when you have a success response.  We keep it for now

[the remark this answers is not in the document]
xlcrr: We also need to start handling false responses

routes/web.php
```diff
@@ -74,6 +74,10 @@
 Route::get('donate/amounts', 'DonateController@index');
 Route::post('donate', 'DonateController@submit');
 
+// Contact page
+Route::get('/contact-us', 'HomeController@index');
+Route::post('/contact-us', 'ContactController')->name('contact');
```
xlcrr, line 79: Should be ContactUsController

app/Mail/ContactMail.php
GeniJaho: We don't need to declare it since it's already declared on the parent Mailable class. It's enough to just set it and it will automatically be handled by the Mailable class to put it as the email's subject.

PR #360, opened 2021-10-18
Disabled exception reporting for incorrect passwords on login

GeniJaho:


xlcrr:
nice

xlcrr:
Can you create a Trello card for this please, and move it to the correct column. Also attach it here and link this PR on the card

PR #364, opened 2021-10-20
Feature/single api for uploads with tags

GeniJaho:
We changed the tags type for this request only. We require tags to be passed as a JSON string instead of an object, due to the differences between multipart/form-data and application/json headers.

The endpoint is `/api/photos/submit-with-tags`


tests/Feature/Api/UploadPhotoTest.php
```diff
@@ -266,10 +266,6 @@ public function test_uploaded_photo_can_have_different_mime_types()
         // JPEG
         $imageAttributes = $this->getImageAndAttributes('jpeg');
         $this->post('/api/photos/submit', $this->getApiImageAttributes($imageAttributes))->assertOk();
-
-        // HEIC
-//        $imageAttributes = $this->getImageAndAttributes('heic');
-//        $this->post('/api/photos/submit', $this->getApiImageAttributes($imageAttributes))->assertOk();
```
xlcrr, line 272: We can add these again?

xlcrr (commented):
Where is the endpoint for the single api upload?

> If we're not going to use the endpoint to store with tags as an object, there's no need to duplicate the endpoint at all. When the need arises to have them both, then we do it.

The problem is people on mobile might not have latest app version, and they need the original endpoint to exist on the backend as they will be uploading objects. We might need to version our API, with v1, v2, etc

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -73,7 +73,7 @@ public function register (Request $request): array
             'username' => 'required|min:3|max:20|unique:users|different:password',
             'email' => 'required|email|max:75|unique:users',
             'password' => 'required|confirmed|min:6|case_diff|numbers|letters',
-            'g-recaptcha-response' => 'required|captcha'
+//            'g-recaptcha-response' => 'required|captcha'
```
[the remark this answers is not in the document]
xlcrr: I think I did this to make the signup work lol

app/Http/Controllers/ApiPhotosController.php
```diff
@@ -289,7 +289,11 @@ public function storeWithTags (Request $request) :array
 
         $userId = Auth::guard('api')->user()->id;
 
-        dispatch (new AddTags($userId, $photo->id, $request->tags));
+        dispatch (new AddTags(
+            $userId,
+            $photo->id,
+            json_decode($request->tags, true)
```
xlcrr, line 295: What if people are using older version of app and do not update it? This will fail?

tests/Feature/Api/UploadPhotoTest.php
GeniJaho: I added them and the tests were failing in the Github Action. That happens because Github servers that run our tests don't have heic support. Plus, every user that wants to contribute to OLM would have to install heic support on their local machine to run tests.

app/Http/Controllers/ApiPhotosController.php
GeniJaho: This is the new route `storeWithTags`, which older app versions don't even know exists, this will be added in the latest versions.

GeniJaho:
> Where is the endpoint for the single api upload?
> 
> > If we're not going to use the endpoint to store with tags as an object, there's no need to duplicate the endpoint at all. When the need arises to have them both, then we do it.
> 
> The problem is people on mobile might not have latest app version, and they need the original endpoint to exist on the backend as they will be uploading objects. We might need to version our API, with v1, v2, etc

We don't delete the old endpoint, it's still functioning the same way as before. On the controller you'll see three methods:
`store`, `storeWithTags`, and `storePhoto` (which is used by the first two as a common operation).  The original `store` method hasn't changed in any way, it's just refactored. On `api.php` we have two routes, the original `/photos/submit` and the new `/photos/submit-with-tags`.

tests/Feature/Api/UploadPhotoTest.php
xlcrr: We should delete them so, maybe add a comment that explains why heic is not tested

app/Http/Controllers/ApiPhotosController.php
xlcrr: I see now, thanks. Comments could be updated when possible

PR #370, opened 2021-11-03
Feature - Picked up litter updates

GeniJaho:
https://trello.com/c/UPqRuLj5

I've added support for a parameter `picked_up` on two API routes for "add tags" and "store photo with tags".
If you pass `picked_up` as `true` the litter will be marked as picked up. Otherwise as still remaining.
If you don't provide this parameter, the picked-up value will default to the user's predefined setting for presence. This is the case for current and older app versions that don't provide this value.

When changing user settings for litter presence, you can now provide a parameter called `picked_up`, which should be true if the litter is by default picked up.

Also, with this PR, every photo retrieved from the backend will have a `picked_up` attribute attached to it, which is just the opposite of `photo->remaining`. Every user instance retrieved will also have a `picked_up` attribute, which is just the opposite of the user setting of `items_remaining`.

I've deleted the parameter type comments on MakeImageAction as they were causing the IDE to scream on controllers, they were more misleading than helping.

This is to be merged after https://github.com/OpenLitterMap/openlittermap-web/pull/364


app/Http/Controllers/ApiPhotosController.php
```diff
@@ -256,7 +256,12 @@ public function addTags (AddTagsApiRequest $request)
             'request' => $request->all()
         ]);
 
-        dispatch (new AddTags($user->id, $photo->id, $request->litter ?? $request->tags));
+        dispatch (new AddTags(
+            $user->id,
+            $photo->id,
+            $request->litter ?? $request->tags,
+            $request->picked_up
```
xlcrr, line 263: what happens if an older app version hits this request and they don't have picked up in the request?

xlcrr (approved):

app/Http/Controllers/ApiPhotosController.php
GeniJaho: If you don't provide this parameter, the picked-up value will default to the user's predefined setting for presence. This is the case for current and older app versions that don't provide this value.

PR #373, opened 2021-11-06
Merge Staging

xlcrr:


xlcrr (commented):
1. `submit-with-tags` should be `upload-with-tags`
2. Teams map not working
3. [Created By label in Settings/Privacy is not working properly](https://trello.com/c/Xwe8hmtG/313-settings-privacy-created-by-section-has-wrong-values)

PR #376, opened 2021-11-13
dd-mm-yyyy-photos.json will now be generated or appended to if it alr…

Xamyrz:
…eady exists while running ```php artisan clusters:generate-all``` first time running might take a while


xlcrr:
I think it might be better to separate this into its own command. Something like `GenerateTextFilesFromAllPhotos`. Then, we also need to update how text files are created + updated after being uploaded from web + mobile

xlcrr:
This PR works but we went with a different approach

PR #385, opened 2021-12-16
Feature - Mobile API for teams

GeniJaho:
https://trello.com/c/kkSWpsZC

This PR adds support for the mobile app to handle teams. It also contains some refactoring for the current TeamsController, since we're doing almost the same actions. Introduced a few FormRequests, and Actions for teams.

TODO on next release:
- download team data
- hide team from leaderboards
- select active  team
- disable active team
- list team members

We expose these endpoints:

```
GET /api/teams/list
	returns array of the user's teams [{}, {}, {}...]


GET /api/teams/types
	returns array of available team types, to be used when creating teams. [{}, {}, {}...]


GET /api/teams/leaderboard
	returns array of top teams, not paginated, no need for authentication. [{}, {}, {}...]


POST /api/teams/create
	parameters: [
		'name' => 'required|min:3|max:100',
        	'identifier' => 'required|min:3|max:15',
	        'team_type' => 'required|integer'
	]

	returns: [
		'team' => The team object
	]

	validation errors: 422
	If user has created the maximum number of teams, you get a 403 Forbidden response, with the appropriate message.



PATCH /api/teams/update/{teamId}
	parameters: [
		'name' => 'required|min:3|max:100',
	        'identifier' => 'required|min:3|max:15',
	]

	returns: [
		'team' => The team object
	]

	validation errors: 422
	If user is not the team leader, you get a 403 Forbidden response.




POST /api/teams/join
	parameters: [
		'identifier' => 'required|min:3|max:15'
	]

	returns: [
		'team' => The team object,
	        'activeTeam' => The user's active team, in case it changes
	]

	validation errors: 422
	If user has already joined the team, you get a 403 Forbidden response.



POST /api/teams/leave
	parameters: [
		'team_id' => 'required'
	]

	returns: [
		'team' => The team object,
	        'activeTeam' => The user's active team, in case it changes
	]

	validation errors: 422
	If user is not a member of the team, or if they're the only member, you get a 403 Forbidden response with the appropriate message.
```


app/Actions/Teams/CreateTeamAction.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace App\Actions\Teams;
+
+use App\Events\TeamCreated;
+use App\Models\Teams\Team;
+use App\Models\User\User;
+
+class CreateTeamAction
+{
+    public function run(User $user, array $data): Team
```
xlcrr, line 11: Comment block missing, add space after run please

app/Http/Controllers/API/TeamsController.php
```diff
@@ -0,0 +1,147 @@
...
+        /** @var User $user */
+        $user = Auth::guard('api')->user();
+
+        if ($user->remaining_teams === 0) {
+            abort(403, 'You have created your maximum number of teams!');
+        }
+
+        $team = $action->run($user, $request->all());
+
+        return ['team' => $team];
```
xlcrr, line 59: I think we should return `'success' => true` with every successful response, it makes our app predictable and easy to check for a success response on the backend

app/Http/Controllers/Teams/TeamsController.php
```diff
@@ -70,61 +73,33 @@ public function inactivateTeam()
...
-        // Broadcast live event to the global map
-        event (new TeamCreated($team->name));
-
-        // Have the user join this team
-        $user->teams()->attach($team);
-
-        $user->active_team = $team->id;
-        $user->remaining_teams--;
-        $user->save();
+        $team = $action->run($user, $request->all());
```
xlcrr, line 83: It's not clear what `$action` is. Maybe we should use the name `$createTeamAction` ?

app/Http/Controllers/Teams/TeamsController.php
```diff
@@ -231,11 +211,11 @@ public function joined ()
      */
     public function members ()
     {
-        $query = $this->filterTeamMembers(request()->team_id);
+        $team = Team::query()->find(request()->team_id);
```
xlcrr, line 214: We can use `Team::find($teamId)` here, no need to use query()

app/Http/Controllers/Teams/TeamsController.php
```diff
@@ -231,11 +211,11 @@ public function joined ()
...
 
-        $total_members = $query->users->count(); // members?
+        $total_members = $team->users->count(); // members?
```
xlcrr, line 216: I think the name `users` is wrong on the team model. It should have a keyword of members. Also, if we are getting the count dynamically, we should drop incrementing and decrementing. In my experience incr/decr rarely works as expected

app/Http/Requests/Teams/UpdateTeamRequest.php
```diff
@@ -0,0 +1,32 @@
...
+                'required',
+                'min:3',
+                'max:100',
+                Rule::unique('teams')->ignore($this->get('id'))
+            ],
+            'identifier' => [
+                'required',
+                'min:3',
+                'max:15',
+                Rule::unique('teams')->ignore($this->get('id'))
```
xlcrr, line 28: Why are we ignoring the ID here? The comment block should explain this. Not immediately clear for the reader

resources/js/store/modules/teams/actions.js
```diff
@@ -352,15 +349,6 @@ export const actions = {
...
             }
-
-            else
-            {
-                Vue.$vToastify.error({
-                    title: failTitle,
-                    body: failBody,
-                    position: 'bottom-right'
-                });
-            }
```
xlcrr, line 363: Why are we removing this team not found by identifier notification? The user will need a response that a team does not exist

routes/api.php
```diff
@@ -90,3 +90,14 @@
...
+
+// Teams
+Route::prefix('/teams')->group(function () {
+    Route::get('/list', 'API\TeamsController@list');
+    Route::post('/create', 'API\TeamsController@create');
+    Route::patch('/update/{team}', 'API\TeamsController@update');
+    Route::post('/join', 'API\TeamsController@join');
+    Route::post('/leave', 'API\TeamsController@leave');
+    Route::get('/types', 'API\TeamsController@types');
+    Route::get('/leaderboard', 'Teams\TeamsLeaderboardController@index');
```
xlcrr, line 102: Can we group the get, post, and other requests together please, and list them A-Z

xlcrr (approved):
Logic looks great, some small changes would make it better

app/Actions/Teams/JoinTeamAction.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace App\Actions\Teams;
+
+use App\Models\Teams\Team;
+use App\Models\User\User;
+
+class JoinTeamAction
+{
+    public function run(User $user, Team $team)
```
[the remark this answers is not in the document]
GeniJaho: Actually, all other big OSS projects refrain from using redundant comment blocks. The argument is that they make the code less readable and potentially confuse the readers, especially when the class/method action is self-explanatory. This is coming from the legends in the Laravel industry https://spatie.be/guidelines/laravel-php#docblocks 
I can add comments for everything from now on, but it would only make the codebase worse, in my opinion (and probably the community's opinion).
Also, let's add a PHP style-ci by the upcoming releases, and unify the whole project's style guide. We'll use https://github.com/FriendsOfPHP/PHP-CS-Fixer

app/Http/Controllers/Teams/TeamsController.php
GeniJaho: The action's purpose is in its type, I don't think it's necessary to use a fully descriptive name here. I kept it short so the method's signature stays in one line.

app/Http/Controllers/Teams/TeamsController.php
GeniJaho: True, but using `query()` allows the IDE to provide type hints and suggestions on the `$team` object, whereas using `find()` directly doesn't, and highlights the following method calls in yellow.

app/Http/Controllers/Teams/TeamsController.php
GeniJaho: True, using `members` would make much more sense. Not sure why we need the `$total_members`, would probably be a good refactoring to just remove it, but we need to make sure it's not used.

app/Http/Requests/Teams/UpdateTeamRequest.php
GeniJaho: You can see the full explanation here https://laravel.com/docs/8.x/validation#rule-unique, on the section `Forcing A Unique Rule To Ignore A Given ID`. This is a very common rule actually.

resources/js/store/modules/teams/actions.js
GeniJaho: I actually changed the existing behavior for the 'team not found' validation. Instead of having it as a separate validation within method code, it's now handled directly in `JoinTeamRequest`. The error message will appear as a validation error, similar to other validation errors.
We use this request both in API and web TeamsController.

routes/api.php
GeniJaho: I've listed them by relevance/importance, but yeah, we can do that 👍

PR #387, opened 2021-12-22
Bugfixes - Small fixes on Global Map and clustering

GeniJaho:
https://trello.com/c/gZ2sSaeP

This PR improves the clustering command:
- Refactoring for more readability
- Use `cursor` instead of `get` to keep the memory consumption low during features generation
- Use Laravel progress bars and console output helpers
- Use batch insert statements instead of single insert per cluster during clusters generation

Clustering is improved at about 3x the speed of the previous version, and there are no memory errors, regardless of total number of photos.

- Schedules the Clustering command to be executed on Sunday at midnight (or Monday at 00:00)
- Added Picked Up status to every photo popup on the global map, teams, profile, and cities maps. Text is translated and is the same as shown on the Tagging page.
- Fixes translation error on Art litter popups, we were using 'arts' as the translation key, but it was 'art' on the translation json. Switched to using 'art' instead.

We need to set the correct .env variable on each server:
`APP_ROOT_DIR=/home/vagrant/code/olm` on local envs,
`APP_ROOT_DIR=/home/forge/olmdev.online` on staging,
`APP_ROOT_DIR=/home/forge/openlittermap.com` on the production server

### DIFF
--- a/app/Models/Litter/LitterCategory.php
+++ b/app/Models/Litter/LitterCategory.php
@@ -43,7 +43,9 @@ abstract class LitterCategory extends Model
         {
             if ($this->$type)
             {
-                $string .= $this->table . '.' . $type . ' ' . $this->$type . ',';
+                $className = $this->table == 'arts' ? 'art' : $this->table;
+
+                $string .= $className . '.' . $type . ' ' . $this->$type . ',';
             }
         }
 
--- a/app/Console/Commands/Clusters/GenerateClusters.php
+++ b/app/Console/Commands/Clusters/GenerateClusters.php
@@ -71,7 +71,7 @@ class GenerateClusters extends Command
         $bar->setFormat('debug');
         $bar->start();
 
-        $photos = Photo::select('lat', 'lon')->get();
+        $photos = Photo::select('lat', 'lon')->cursor();
 
         $features = [];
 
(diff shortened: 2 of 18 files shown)


app/Models/Litter/LitterCategory.php
```diff
@@ -43,7 +43,9 @@ public function translate ()
         {
             if ($this->$type)
             {
-                $string .= $this->table . '.' . $type . ' ' . $this->$type . ',';
+                $className = $this->table == 'arts' ? 'art' : $this->table;
+
+                $string .= $className . '.' . $type . ' ' . $this->$type . ',';
```
xlcrr, line 48: String interpolation would be better here 

```
$string .= "$className.$type.$this->type,";
```

Also, it helps to evaluate expression with ()

```
$x = ($this->table === 'arts') ? 'art' : $this->table;
```

xlcrr (approved):
Looks really good! Love the new clustering command

app/Models/Litter/LitterCategory.php
GeniJaho: This was actually a quick fix more than a proper solution. In the future, we should switch from using the table name to using the snake_cased class name, it's probably better.

PR #392, opened 2022-01-07
Feature - Yearly global clustering

GeniJaho:
https://trello.com/c/gZ2sSaeP

- Added a `year` column to the `clusters` table, with a proper MySql index
- Also added an index to the `photos.datetime` column, so that filtering by `datetime` will be much faster. It's weird however that this column's type is `varchar`. It should be `datetime`, we're losing so much performance this way. Is there any chance we can fix that in a future release?
- We should run `php artisan migrate` when deploying this feature.
- Updated the global clustering command to take years into account, starting from 2017
- Updated the endpoints to filter data (clusters and points) by year. We do not do any filtering for Art data.

In the front-end:
- Copied the code from this package https://github.com/adammertel/Leaflet.Control.Select, there was no other way to use it by importing it through npm. I left a comment on the copied files with the original link.
- Also, the only way to make this work is by reloading the page when the user selects a year. It's the cleanest possible way to implement changing the data.
- Since we're using the `year` parameter on the URL, now we can also share the map for a particular year, like this: http://olm.test/global?year=2021&lat=54.44229334400231&lon=29.02682479541212&zoom=4.07
- By default, we show the clusters for `All Time`

For the time being, I suggest we cluster every day, all the years. At least until we have confirmation from our users that they have uploaded all their old photos.

The result looks like this:
image


database/migrations/2022_01_05_151733_add_year_to_global_clusters.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('clusters', function (Blueprint $table) {
+            $table->year('year')->nullable(true)->index();
```
xlcrr, line 17: I don't think we need to call nullable = true? I think we can just use `nullable()`

database/migrations/2022_01_05_151733_add_year_to_global_clusters.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('clusters', function (Blueprint $table) {
+            //
```
xlcrr, line 35: We need to include reverse migrations

xlcrr (changes requested):
Need to test how the dropdown looks. Can you include a screenshot? We had a custom component for this in a previous release. Dropdown dates I think

app/Http/Controllers/GlobalMap/ClusterController.php
```diff
@@ -16,11 +17,12 @@ class ClusterController extends Controller
     /**
      * Get clusters for the global map
      *
+     * @param Request $request
      * @return array
      */
-    public function index(): array
+    public function index(Request $request): array
```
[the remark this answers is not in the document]
GeniJaho: I actually included the Request variable here to skip redundant calls to request(). They're basically the same object, there are really no differences in which you use. Just that request() goes a long way into the container to fetch the object.

database/migrations/2022_01_05_151733_add_year_to_global_clusters.php
```diff
@@ -0,0 +1,40 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('clusters', function (Blueprint $table) {
+            $table->year('year')->nullable()->index();
+
+            $table->index(['year', 'zoom']);
```
xlcrr, line 19: Do we need to index year again here? Is it doing an index combination of year + zoom?

xlcrr (changes requested):

database/migrations/2022_01_05_151733_add_year_to_global_clusters.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('clusters', function (Blueprint $table) {
+            Schema::table('clusters', function (Blueprint $table) {
```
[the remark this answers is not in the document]
GeniJaho: Yes, dummy mistake 😅 Pushed the fix.

database/migrations/2022_01_05_151733_add_year_to_global_clusters.php
GeniJaho: Yes, it's doing a composite index. In our queries, we will always be doing queries with both year and zoom, even when the year is null.

xlcrr (approved):

PR #396, opened 2022-01-13
Bugfix - API upload photo multiple times

GeniJaho:
https://trello.com/c/uHEEAf0f

### DIFF
--- a/app/Http/Controllers/ApiPhotosController.php
+++ b/app/Http/Controllers/ApiPhotosController.php
@@ -9,6 +9,8 @@ use App\Actions\Photos\UploadPhotoAction;
 use App\Events\ImageDeleted;
 use App\Http\Requests\AddTagsApiRequest;
 use App\Models\Photo;
+use App\Models\User\User;
+use Exception;
 use GeoHash;
 use Carbon\Carbon;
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ApiPhotosController.php
```diff
@@ -148,6 +151,12 @@ protected function storePhoto(Request $request): Photo
 
         $date = Carbon::parse($date);
 
+        if (app()->environment() === "production") {
+            if (Photo::where(['user_id' => $user->id, 'datetime' => $date])->exists()) {
+                throw new Exception("You have already uploaded this file!");
```
xlcrr, line 156: should we return something more generic to the mobile app, like ['success' => false, msg => 'exists']` ?

xlcrr: Also, please allow user_id = 1 to duplicate images on production, sometimes I need this for testing

xlcrr (approved):

xlcrr:
@GeniJaho we can merge this but I need user.id 1 to duplicate photos on production for testing

app/Http/Controllers/ApiPhotosController.php
GeniJaho: The mobile app does not need the 'success' => false info, as the request status will be 500 or 422. The message is already there. What do we get with this current implementation on mobile?

I'm allowing the user_id=1 on the next bugfix PR.

GeniJaho: I checked on mobile, nothing shows to the user, just that Something went wrong. Need to discuss this with Anoop to see what he needs on his side.

Issue #97, opened 2020-12-17
Credit the creator of a city with his name (created by: anonymous --> created by: 'realname')


FredMark:
If you go to a province, for instance The Netherlands > North Holland,
you see the different cities that have been created because litter was picked in them.
At this moment you see 'created by: Anonymous'.

If you go to 'settings > Privacy', indeed it mentions that:
"Your name and username will not appear in the Created By section of any locations you add to the database."

Probably something on the todo-list, but could be nice if this was working.
(where people can choose with what name they want to be mentioned: name or username (which is 'unique identifier' in personal-details-screen?) or maybe even 'none'

xlcrr:
This is working for me on production, maybe after the merge

FredMark:
As discussed in Slack,
please don't close issues merely because 'it is working on my machine',
This one doesn't seem to be solved but i never will know that it is closed unless by coincidence.
Please inform me if something seems to be working (that is, if the issue was raised by me) so i can see if i can conclude the same.
In this case, i still have the same issue, some something else seems to be the case so i re-opened it.

xlcrr:
Not working for me either... will fix soon

PR #398, opened 2022-01-13
Feature- Admin filter photos

GeniJaho:
https://trello.com/c/3sI94kHi
https://trello.com/c/8nQnguJT

- Added filtering photos on Admin Verify Photos page by Country
- Added option to Skip photos
- Removed some informative texts and put them into button tooltips.
- Refactored RecentTags into its own component, and pushed it on the lower left side of the screen.
- When you click on Verify/Delete/False etc on the admin Verify Photos page, and you load the next picture, the page will now scroll on top, not at the bottom of the screen where the user might have scrolled.
- Added 'auth' middleware for all team routes, and fixed a small bug that prevents you from updating your team name/username separately.
- Fixed DeletePhotoAction errors we've received on production. If somehow a photo does not have a filename or a bbox filename, we simply do nothing since the photo doesn't exist. We don't throw an exception since it might have been a partial upload.
- Fixed the Creator of a city/state/country issue showing Anonymous instead of their real name (https://github.com/OpenLitterMap/openlittermap-web/issues/97)

image


app/Http/Controllers/AdminController.php
```diff
@@ -228,50 +232,98 @@ public function updateTags (Request $request)
...
+        $totalsQuery = Photo::query()
+            ->selectRaw('country_id, count(*) as total')
+            ->whereIn('verification', [0, 0.1])
+            ->whereNotIn('user_id', $this->usersToSkipVerification())
+            ->groupBy('country_id');
+
+        // Using DB to avoid extra appended properties
+        return DB::table('countries')
+            ->selectRaw('id, country, q.total')
+            ->rightJoinSub($totalsQuery, 'q', 'countries.id', '=', 'q.country_id')
```
xlcrr, line 302: what is `q` ? query I guess..

app/Http/Controllers/PhotosController.php
```diff
@@ -186,7 +181,7 @@ public function store (Request $request)
         $latitude = $latlong[0];
         $longitude = $latlong[1];
 
-        $revGeoCode = $this->reverseGeocodeAction->run($latitude, $longitude);
+        $revGeoCode = app(ReverseGeocodeLocationAction::class)->run($latitude, $longitude);
```
xlcrr, line 184: any reason why this changed?

resources/js/store/modules/admin/actions.js
```diff
@@ -133,10 +144,27 @@ export const actions = {
...
             .catch(err => {
                 console.error(err);
             });
-    }
+    },
 
+    /**
+     * Get list of countries that contain photos for verification
+     */
+    async GET_COUNTRIES_WITH_PHOTOS (context)
```
xlcrr, line 158: I think all admin action names should be prefixed with `ADMIN_` to make it clear this is not a normal user operation

xlcrr (approved):

app/Http/Controllers/AdminController.php
GeniJaho: Yes, every subquery needs an alias, 'q' is quite common.

app/Http/Controllers/PhotosController.php
GeniJaho: Yes, I added a fake for this action, which allows me more options to properly mock the geocoding action. I don't need to mock this class in every test anymore. Using dependency injection without an interface forces me to use app() in this case. Nothing wrong with it though, this code piece is perfectly fine.

PR #400, opened 2022-01-19
Prevent token from being null

xlcrr:
Some users are saying that verify email token results in 404 

The first time I click the link it works, but additional clicks result in user not found 

Maybe we should keep the token and let them press it again ?

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -97,7 +97,10 @@ class RegisterController extends Controller
     */
     public function confirmEmail ($token)
     {
-        $verified = User::whereToken($token)->firstOrFail()->confirmEmail();
+        /** @var User $user */
+        $user = User::whereToken($token)->first();
+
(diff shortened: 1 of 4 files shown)


GeniJaho:
Why do they need to click the link twice? If they click it they'll get verified. The link should return a 404, to indicate that this action cannot be done anymore, as they're already verified. I think they're clicking it to just get to the home page, without typing the URL in the browser 😅

xlcrr:
I'm not sure, but many users have reported this bug. I think we can just let them re-verify their email, its not that important

GeniJaho:
If they're verified, we can probably just return them to the home page, not showing a 404. But the token should expire I think.

xlcrr:
That makes sense to me too, just as long as we are not sending new users to a 404, it looks broken

GeniJaho (approved):

GeniJaho:
Added a simple check to just send the users to the homepage, if their token is already used. We can merge this.

PR #401, opened 2022-01-24
Bugfix - Teams mobile API

GeniJaho:
- Fixed teams mobile api authentication
- Fixed mobile api duplicate image response
- Prefixed admin action with ADMIN_

### DIFF
--- a/app/Http/Controllers/API/TeamsController.php
+++ b/app/Http/Controllers/API/TeamsController.php
@@ -22,7 +22,7 @@ class TeamsController extends Controller
 
     public function __construct()
     {
-        $this->middleware('auth')->except('types');
+        $this->middleware('auth:api')->except('types');
     }
 
     /**
@@ -33,7 +33,7 @@ class TeamsController extends Controller
     public function list()
     {
         /** @var User $user */
-        $user = Auth::guard('api')->user();
+        $user = Auth::user();
 
         return $user->teams;
     }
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/API/TeamsController.php
```diff
@@ -33,7 +33,7 @@ public function __construct()
     public function list()
     {
         /** @var User $user */
-        $user = Auth::guard('api')->user();
+        $user = Auth::user();
```
xlcrr, line 36: I'm not sure if this works with api, we will see

app/Http/Controllers/ApiPhotosController.php
```diff
@@ -146,9 +146,12 @@ protected function storePhoto(Request $request): Photo
 
         $date = Carbon::parse($date);
 
-        if (app()->environment() === "production") {
+        // The user with id=1 needs to upload duplicate images for testing
+        if (app()->environment() === "production" && $user->id != 1) {
             if (Photo::where(['user_id' => $user->id, 'datetime' => $date])->exists()) {
-                throw new Exception("You have already uploaded this file!");
+                return response()->json(['errors' => [
+                    'photo' => 'You have already uploaded this file!'
```
xlcrr, line 153: a simple `'already exists'` might do

xlcrr (approved):

app/Http/Controllers/API/TeamsController.php
GeniJaho: The tests are passing for this change. The thing is that when using the 'auth:api' middleware, the framework uses the 'api' guard by default. Anoop should confirm this though.

PR #411, opened 2022-02-01
Small teams API refactoring

GeniJaho:
https://trello.com/c/GZIk6vMN

Simply moves the check for team membership into the User model.
Also fixes the issue of clicking on a recent tag and incrementing the quantity.

### DIFF
--- a/app/Http/Controllers/API/TeamsController.php
+++ b/app/Http/Controllers/API/TeamsController.php
@@ -96,7 +96,7 @@ class TeamsController extends Controller
         $team = Team::whereIdentifier($request->identifier)->first();
 
         // Check the user is not already in the team
-        if ($user->teams()->whereTeamId($team->id)->exists()) {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/API/TeamsController.php
```diff
@@ -179,9 +179,9 @@ public function members(ListTeamMembersAction $action): array
         /** @var User $user */
         $user = auth()->user();
         /** @var Team $team */
-        $team = Team::query()->findOrFail(request()->team_id);
+        $team = Team::query()->find(request()->team_id);
```
xlcrr, line 182: After this line, we should do 
 ```
 if (!$team) {
     return [
         'success' => false,
         'msg' => 'not found'
     ];
 }
 ```

xlcrr: or 
```
$this->fail('not-found');
```

xlcrr (approved):

PR #417, opened 2022-02-08
Fixed litter art text in the Global Map

GeniJaho:
https://trello.com/c/velI39D4

We simply didn't pass the art items `result_string` to the rendering method. We pass it now.
I changed the text to read 'Litter Art', without the quantity of 1.

I refactored the command that generates the result string when an image is tagged. We could improve further by compiling the result string of only the photo being tagged, but that's for another time.

image

### DIFF
--- a/app/Console/Commands/CompileResultsString.php
+++ b/app/Console/Commands/CompileResultsString.php
@@ -43,14 +43,11 @@ class CompileResultsString extends Command
      */
     public function handle()
     {
-        $photos = Photo::where('verified', '>', 0)->get();
-
-        foreach ($photos as $photo)
-        {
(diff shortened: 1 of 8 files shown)


app/Console/Commands/CompileResultsString.php
```diff
@@ -43,14 +43,11 @@ public function __construct()
...
-            if (is_null($photo->result_string))
-            {
-                $photo->translate();
-            }
-        }
+        Photo::query()
+            ->where('verified', '>', 0)
+            ->whereNull('result_string')
+            ->get()
+            ->each
```
xlcrr, line 50: does this need `()` ?

xlcrr (approved):

PR #424, opened 2022-02-21
Feature - Community page

GeniJaho:
https://trello.com/c/IKXmVH97


resources/js/views/home/Community/StatsSection.vue
```diff
@@ -0,0 +1,182 @@
...
+                ]
+            };
+        },
+
+        /**
+         * Format number value
+         */
+        commas (n)
+        {
+            return parseInt(n).toLocaleString();
```
xlcrr, line 118: We actually have a global filter helper for this on app.js, you can do 

```
{{ n | 'commas' }}
```

xlcrr (approved):

resources/js/views/home/Community/StatsSection.vue
GeniJaho: True, but we're using the `<number />` component above, hence the method declared here.

PR #433, opened 2022-03-03
Bugfix - Fix picked up on global map

GeniJaho:
https://trello.com/c/4MrINi2b
https://trello.com/c/KtCqRTrq

The issue was that when getting the features on the GlobalMapController, we didn't include the 'remaining' column, so the 'picked up' value always defaulted to true.

I also removed the last month from the stats on the Community page, since it's always 0.

We also skip the 'All Time' clustering if there are no new photos uploaded recently for any year.

### DIFF
--- a/app/Http/Controllers/PhotosController.php
+++ b/app/Http/Controllers/PhotosController.php
@@ -333,7 +333,7 @@ class PhotosController extends Controller
 
         $this->updateLeaderboardsAction->run($photo, $user->id, $litterTotals['all'] + $customTagsTotal);
 
-        $photo->remaining = !$request->presence;
+        $photo->remaining = !$request->picked_up;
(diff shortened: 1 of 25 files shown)


app/Http/Controllers/PhotosController.php
```diff
@@ -333,7 +333,7 @@ public function addTags (AddTagsRequest $request, AddCustomTagsToPhotoAction $cu
 
         $this->updateLeaderboardsAction->run($photo, $user->id, $litterTotals['all'] + $customTagsTotal);
 
-        $photo->remaining = !$request->presence;
+        $photo->remaining = !$request->picked_up;
```
xlcrr, line 336: need to double check what value is being passed here. Would be good to update the column name to picked up...

xlcrr (approved):
ignore comment

PR #450, opened 2022-04-10
Add two new brands

GeniJaho:
https://trello.com/c/h3IMsNGn
https://trello.com/c/93XqEbpL

Added Wendy's and Tim Hortons brands.
You should compile for production as well.


database/migrations/2022_04_10_210041_add_new_brand_tags2.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
xlcrr, line 29: missing reverse migrations?

xlcrr (approved):

database/migrations/2022_04_10_210041_add_new_brand_tags2.php
GeniJaho: We shouldn't be doing reverse migrations anyway, it's more stable like this. I've learned this from others, and the hard way as well, lol.

PR #451, opened 2022-04-11
Bugfix - Tagging issues & Global Map fixes

GeniJaho:
https://trello.com/c/lSyo7OcA
https://trello.com/c/oCNDhlyI
https://trello.com/c/kaig1wB2
https://trello.com/c/HX0m6QzW
https://trello.com/c/qrjMVmpC
https://trello.com/c/cL0RlrUT
https://trello.com/c/ckDPF6Y5
https://trello.com/c/1bDR7CfV
https://trello.com/c/kLc0aynC

Next round of Tagging page improvements:
- When clicking a Recent Tag we set it as selected so you can now change the quantity easily.
- When clicking a Recent Tag it will now move to the top of the list.
- Made the Recent Tag box a bit longer, and swapped the Photo details box with the Items Remaining box (the photo details now are on the right, the photo below).
- Rounded the corners of the image, same radius as the other boxes in the screen.
- Fixed the duplicate custom tags errors that appeared on production.
- Cleared custom tags errors when switching pages
- This is already on production, but you can actually autocomplete the search results when tagging by pressing Ctrl/Shift + Spacebar. It would be nice to show that to the users with a tooltip, maybe in the next release.
- Left-aligned search results and tag lists, so that they're more readable and distinguishable.

Global Map
- When opening a photo location, we don't wait for 3 seconds to open it anymore, whenever the photo is ready we load it.
- When we click on a point, we update the URL with the photo id
- When we zoom out from points to clusters, remove the photo id
- When clicking on the easter egg, we update the URL with the photo id
- When double-clicking the easter egg we don't hide the event anymore
- Users can now apply filters in the URL to filter out by date or user. You can provide them like this: `fromDate=2017-02-25&toDate=2018-01-01&username=geni`. Filters are independent of each other, so we can apply only fromDate, or only username, if we want. Also, a user needs to have their 'show_username_maps' setting checked in order for `username` filtering to work.

Not implemented:
- Instead of opening a photo link in a new tab, copy it. This is actually pretty hard to do from the way we render that popup, there's a detailed comment here https://trello.com/c/HX0m6QzW.
- When changing pages, don't clear the array of events. This was buggy as well, somehow the page wouldn't listen for new events anymore, detailed comment here https://trello.com/c/1bDR7CfV.

New Tagging page:

image


resources/js/components/LiveEvents.vue
```diff
@@ -123,10 +122,34 @@ export default {
...
 	},
 	methods: {
+        /**
+         * This is usually how double-clicks are handled
+         * without overlapping with the click events
+         * @see https://stackoverflow.com/a/41309853/5828796
+         * @param event
+         * @param index
+         */
+        click(event, index)
```
xlcrr, line 138: Vue also has a `@dbClick` handler

GeniJaho: That's what we have currently on production, and that works. This issue is, however, that it does not prevent the 'click' event from being fired as well. All solutions were like this.

xlcrr (approved):

PR #463, opened 2022-04-27
Feature - Social links & Profile page fixes

GeniJaho:
https://trello.com/c/DwYSbXx6
https://trello.com/c/EvNxP2Hu
https://trello.com/c/1MJylEjW
https://trello.com/c/Z8Zf52E7

Added the ability to show your social accounts:
- New section in Settings called Social Accounts, you can provide URLs for Twitter, Facebook, Instagram, Reddit, LinkedIn, and your personal site (or whatever URL you want).
- These links will be visible on every picture you upload on the Global Map, regardless of your privacy settings. These links will also appear on the Global Leaderboard (top 10 users) when you hover over the users.

image
image

Fixed the map in the Profile Page
- We don't use clusters anymore, that piece of functionality was broken.
- Instead, we use the same technique as in the City map, just without the hex grids. Filters are working. After this we might want to refactor the UI to have only one place for filters that can be used for both the Map and the Downloads.

When a new country, state, or city is added, we send a notification to slack, the `#olm-alerts` channel.
You need to provide a `SLACK_WEBHOOK_URL=` configuration in the staging and production servers.
The channel can be configured through the `SLACK_WEBHOOK_CHANNEL=` env configuration. 

Reverted the changes on Recent Tags. When you click on a recent tag it won't animate anymore, and when adding a new recent tag it will be added to the bottom and to the right of the list, like before.

Moved the 'Other' category to the bottom of the list in the Tagging/Verification screen.

Refactored the mapHelpers.js to require fewer parameters, it's easier now to use it.


xlcrr:
Todo
1. Icons should be in their native colour
2. Only show the icons that the user has connected
3. By default, no icons show. User needs a setting to toggle icons on or off.
4. Setting needs to be on icon needs a link for each icon to show
5. User can save their links, and turn the setting off to hide icons

GeniJaho:
Hmm, but:
1. What's wrong with the link color? We're using font-awesome for easy access to icons. Would it really make a difference to find icons of each type, resize and stretch them? Plus, we can control their color like this, white for black backgrounds, and vice-versa.
2. We're only showing icons when the user provides a URL for it.
3. Not showing any icons by default, because no user will have those links set up. 
4. and 5. The users simply don't fill the social networks they don't want. Why would they provide a link and keep it hidden? 
We also have a descriptive text in the Social Accounts section
"These links will appear in every photo you've uploaded, on the Global Map. Please provide full URLs."

PR #464, opened 2022-04-29
Global leaderboard for all users

xlcrr:
[Trello card](https://trello.com/c/Ce1wJLhV/422-turn-global-leaderboard-into-a-link-that-paginates-entire-leaderboard-by-100-users)

Added our own custom pagination solution for getting the data from redis


app/Http/Controllers/Leaderboard/LeaderboardController.php
```diff
@@ -0,0 +1,69 @@
...
+            $start = ($currentPage * $perPage) + 1;
+
+            // 200
+            $end = $end + ($currentPage * $perPage);
+        }
+
+        $userIds = Redis::zrevrange("xp.users", $start, $end);
+
+        $users = User::whereIn('id', $userIds)
+            ->select('id', 'show_name', 'show_username', 'name', 'username')
```
xlcrr, line 37: todo - add socials

resources/js/components/global/GlobalLeaders.vue
```diff
@@ -26,40 +32,65 @@
...
+                    v-show="currentPage > 0"
+                    class="button is-medium mr-1"
+                    @click="loadPreviousPage"
+                >Previous</button>
+
+                <button
+                    v-show="leaderboard.users.length === 100"
+                    class="button is-medium"
+                    @click="loadNextPage"
+                >Next</button>
```
xlcrr, line 61: todo - add translations

app/Http/Controllers/Leaderboard/LeaderboardController.php
GeniJaho: Actually, we're better off just showing all users according to their XP. If they don't want their name shown, then we just show their name as Anonymous. That's similar to what we do in our existing country, state, and city leaderboards. This will make the code simpler because there's just too much duplication here.

GeniJaho: And will make it easier to include socials here.

GeniJaho (commented):
@xlcrr I can take it from here if you want.

PR #466, opened 2022-05-02
Merge staging

xlcrr:


GeniJaho:
Before we merge, or maybe after 👀 we should add the translations for all new and updated texts.
Also, we've run npm run dev for this, we should run prod.

PR #476, opened 2022-05-09
Fix deployment

GeniJaho:
This PR removes the public/app.js from gitignore, and commits the file into version control.
I've also fixed a failing test for teams, we're showing only 2 team members instead of 10 per page. Reverted it back to 10.


resources/js/components/global/GlobalLeaders.vue
```diff
@@ -7,14 +7,15 @@
 						<th class="has-text-centered" style="width: 15%;">{{ $t('location.position') }}</th>
 						<th>{{ $t('location.name') }}</th>
 						<th class="has-text-centered">{{ $t('location.xp') }}</th>
-                        <th class="has-text-centered hide-mobile">Socials</th>
+                        <th class="has-text-centered hide-mobile">Social</th>
```
xlcrr, line 10: This was not a bug :D

GeniJaho: I didn't do this change, it's coming from your commits on master. lol

PR #489, opened 2022-06-01
Bugfix - Routine fixes 5, Material category

GeniJaho:
This PR adds a new Material category with these tags, no translations yet
(https://trello.com/c/ONonEnVD, https://trello.com/c/r1ncEbc5):
```
- Iron/Steel
- Aluminium
- Copper
- Titanium
- Nickel
- Bronze
- Metal

- Concrete
- Fiberglass
- Carbon Fiber
- Composite

- Glass
- Ceramic

- Rubber
- Latex
- Polyethylene
- Polypropylene
- Polystyrene
- PVC
- Nylon
- Plastic
- Polymer

- Paper
- Wood
```
They are sorted alphabetically in the actual implementation.

Other changes & fixes include:
- Suggest your recent custom tags when searching for tags on the Tagging page. You can select them the same way as regular tags. If what you search for does not exist, you still have to type it in the custom-tag field to add a new custom tag. (https://trello.com/c/UXzKzdMD)
- Removed 'Code=1' and 'Code=2' from error messages when uploading images with no GPS location.
- Updated the min width and height of map image popups, hopefully, this time they'll be easy to navigate around and show a larger part of the images.
- Made the font size a bit smaller on mobile for the map popups.
- Made the font size a bit smaller for live events, it looks more compact now.
- On live events, we show either 'Litter Picked Up' or 'Litter Uploaded', regardless of the user verification status.
- On live events, made the country flags to 'fill' instead of 'cover'. This fixes some flags, like Ireland's, but ruins and stretches some others, like Albania's 😂. Tradeoffs.
- Fixed a bug when submitting tags via the keyboard shortcut (we registered the event listener on every photo but didn't unregister it).


xlcrr:
```
On live events, made the country flags to 'fill' instead of 'cover'. This fixes some flags, like Ireland's, but ruins and stretches some others, like Albania's 😂. Tradeoffs.
```

I noticed this too, we might have to choose the correct class for each location 1 by 1...

app/Models/Litter/Categories/Material.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace App\Models\Litter\Categories;
+
+use App\Models\Litter\LitterCategory;
+
+class Material extends LitterCategory
+{
+    protected $table = 'material';
```
xlcrr, line 9: Do these need to be mass assignable?

xlcrr (approved):

app/Models/Litter/Categories/Material.php
GeniJaho: Yep, I think we do it on the parent model.

PR #492, opened 2022-06-02
Fixed custom tags suggestions name collision

GeniJaho:
Fixes the errors when adding a custom tag that already has a regular tag with the same key.
Simply updates the custom tag keys with the string 'custom-' so that there are no collisions with the regular tags.


xlcrr:
Nice catch

PR #493, opened 2022-06-07
Feature- Convert Deprecated Tags

GeniJaho:
https://trello.com/c/BmjgFbeY

I've implemented the conversions in this [document](https://docs.google.com/spreadsheets/d/1wRcojHGgl0KQA85cg5g2mr2o1z4_Z2wnmHLiMFQKEj4/edit).

- Some tags are disabled, and instead, we've added new ones.
- Added translations for the new tags and the Material category tags.
- We've added a simple conversion mechanism to support the old or current mobile apps that will continue to submit the old tags, shown in the document above. I've written tests to make sure it works as expected.
- I 've also added a new console command that unifies the translation files with the English version, it simply adds the missing keys to the other languages.


database/migrations/2022_06_06_220223_add_new_tags_v2.php
```diff
@@ -0,0 +1,49 @@
...
+            $table->integer('cup')->unsigned()->nullable();
+            $table->integer('packaging')->unsigned()->nullable();
+        });
+        Schema::table('coastal', function (Blueprint $table) {
+            $table->integer('degraded_bottle')->unsigned()->nullable();
+            $table->integer('degraded_bag')->unsigned()->nullable();
+        });
+        Schema::table('food', function (Blueprint $table) {
+            $table->integer('packaging')->unsigned()->nullable();
+            $table->integer('cutlery')->unsigned()->nullable();
```
xlcrr, line 26: These new additions are welcome, but I think we should keep the original ones too, as it enables the user to select the material + item at once. Will see what the group thinks

PR #497, opened 2022-06-16
Bugfix - Routine fixes 7

GeniJaho:


app/Models/Photo.php
```diff
@@ -285,4 +287,13 @@ public function customTags(): HasMany
     {
         return $this->hasMany(CustomTag::class);
     }
+
+    public function scopeOnlyFromUsersThatAllowTagging(Builder $query)
+    {
+        $query->whereNotIn('user_id', function ($q) {
+            $q->select('id')
+                ->from('users')
+                ->where('prevent_others_tagging_my_photos', true);
```
xlcrr, line 296: We want to get data where `prevent_others_tagging_my_photos` is false?

A clearer name would be `allow_others_tag_my_photos`?

xlcrr (approved):

app/Models/Photo.php
GeniJaho: It's not a bug, we filter out these users, we have a test for this. It's called like that because the field will be null for existing users.

PR #510, opened 2022-07-18
Admin: Auto-verify user after first Littercoin mined

xlcrr:


GeniJaho (commented):
I can't really review the changes here since I don't know the approach, looks ok in general. I think we need to fix the few failing tests though, or just delete them if they've become irrelevant and we don't longer need what they're testing.
Why did we migrate to Redis though? There were no performance issues here.

xlcrr:
I am going to move any incrementing data to Redis. Database will be a source of truth, redis will be the metadata that makes representing the database easier to manage

PR #527, opened 2022-09-08
French translation

Brautantoine:
Hi, I love the concept so I tought about tying to contribute a little

I try to provide a french translation. I'm not too familiar with OpenMapLitter yet (I've discovered it recently) so I hope I didn't made  "quirky" translations.

Also I think "étiquettes" would be better than "Tags", but tags was already used in the previous translations so I stick with it


xlcrr (approved):

xlcrr:
Excellent work! Thanks so much

PR #532, opened 2022-09-18
Update - Allow user to edit their submitted photos

xlcrr:
Work in progress...


GeniJaho:
PR looks good to me Sean, I think you really want to add some tests for the main features you're introducing. I'm sure you'll find a bug or two while adding tests 😅 
You could test the DB integrity and everything that might be affected by this.

PR #534, opened 2022-09-18
Dump the database schema

GeniJaho:
This schema dump will make the tests faster since they'll no longer have to run the migrations one by one.
On my lame PC, the execution went from 2 min 17 secs to 1 min 52 seconds, a 25s decrease. It might have a bigger impact on a good Mac.

https://laravel.com/docs/8.x/migrations#squashing-migrations

It has no effect on the existing database, and nothing changes when running migrations.
You can try running this locally and confirm that nothing happens if you run 'php artisan migrate'.

### DIFF
--- a/database/schema/mysql-schema.dump
+++ b/database/schema/mysql-schema.dump
@@ -1,0 +1,1837 @@
+/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
+/*!40103 SET TIME_ZONE='+00:00' */;
+/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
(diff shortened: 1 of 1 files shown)


xlcrr:
Nice find! I thought we had to upgrade to a later version of Laravel to achieve this. How do we know its not going to accidentally delete the database when it merges into production? I also ran this locally, didn't notice any speed improvement with `php artisan test`

GeniJaho:
> Nice find! I thought we had to upgrade to a later version of Laravel to achieve this. How do we know its not going to accidentally delete the database when it merges into production? I also ran this locally, didn't notice any speed improvement with `php artisan test`

Maybe it's always going to be fast on a Mac lol.
Actually, we don't need this at all. I noticed we can get away with squashing only locally to improve the tests.
Closing the PR.

PR #559, opened 2022-12-08
Feature - Filter by multiple custom tags

GeniJaho:
Allows you to filter photos by multiple custom tags at once, like this `/tags-search?custom_tags=tag_1,tag_2,tag_3`.
Photos must have all of the tags listed to be included in the results.


xlcrr:
Merged this into staging, but it doesn't seem to be working?

https://olmdev.online/tags?custom_tags=asd,fgh

`asd` exists, `fgh` does not. All data is loaded

GeniJaho:
Oh I've only implemented the change for the back end request to /tags-search, but I think we need to do a small update in the front-end component as well.

PR #585, opened 2023-03-17
Feature: Littercoin Integration

lley154:
- added remaining transactions (burn, add ada, mint merchant token)
- updated UI including spinner
- renamed node.js scripts for clarity
- imporved error handling routines

Next steps is to do final code cleanup and complete testing


app/Http/Controllers/Littercoin/LittercoinController.php
```diff
@@ -24,16 +33,255 @@ public function getUsersLittercoin ()
...
+        $response = exec($cmd);
+        $responseJSON = json_decode($response, false);
+
+        if ($responseJSON->status == 200) {
+
+            // Update the amount of littercoin paid to user in the DB
+
+            // Commented out for testing purposes
+            //$user->littercoin_paid = $littercoinPaid + $littercoinDue;
+            //$user->save();
```
xlcrr, line 105: We should probably update each littercoin in the new table as sent

app/Http/Controllers/Littercoin/LittercoinController.php
```diff
@@ -24,16 +32,270 @@ public function getUsersLittercoin ()
...
+     */
+    public function mintTx (Request $request) {
+
+        $destAddr = $request->input('destAddr');
+        $changeAddr = $request->input('changeAddr');
+        $utxos = $request->input('utxos');
+        $strUtxos=implode(",",$utxos);
+
+        $userId = Auth::user()->id;
+        $littercoinPaid = Auth::user()->littercoin_paid;
```
xlcrr, line 61: I think we should calculate this from the littercoin table and update each littercoin as sent or not

lley154: agreed, with the latest commits, we are now only reading and update the littercoin table

app/Http/Controllers/Littercoin/LittercoinController.php
lley154: this has been completed in recent commits

xlcrr (approved):

PR #592, opened 2023-04-07
updates to about page

SmashedFinger:
This commit came from the github.dev website, and was possible without setting up the environment.


Newman5:
I was there when this happened.  Seems legit

PR #608, opened 2023-04-28
small text update to littercoin page

SmashedFinger:


SmashedFinger:
Not sure I did this correctly. I just updated a line of text from the littercoin page. tried to merge directly into master but got an error so tried staging instead.

PR #630, opened 2023-07-10
Backend API to Delete account + all records

xlcrr:


app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+
+                            if (sizeof($photo->customTags) > 0)
+                            {
+                                foreach ($photo->customTags as $customTag)
+                                {
+                                    $customTag->delete();
+                                }
+                            }
+
+                            $adminVerificationLogs = AdminVerificationLog::where([
```
GeniJaho, line 54: You don't need to do the count for verification logs, just delete them.

app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+                                'user_id' => $userId,
+                                'photo_id' => $photo->id
+                            ])->count();
+
+                            if ($adminVerificationLogs > 0)
```
GeniJaho, line 59: You don't need all the other checks, it's safe to delete the rows directly.

app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+                        }
+                    }
+                }
+            });
+
+        try
+        {
+            $adminVerificationLogs = AdminVerificationLog::where('user_id', $userId)->count();
+
+            if ($adminVerificationLogs > 0)
```
GeniJaho, line 110: You don't need these checks.

app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+
+            $cleanups = Cleanup::where('user_id', $userId)->count();
+
+            if ($cleanups > 0)
+            {
+                Cleanup::where('user_id', $userId)->delete();
+            }
+
+            $countries = Country::select('id', 'created_by', 'user_id_last_uploaded')
+                ->where('created_by', $userId)
```
GeniJaho, line 130: You need to split this into two queries since a country/state/city might have been created by the user to be deleted, but the user_id_last_uploaded could belong to another user, and we're setting it to null regardless. And we don't need the checks.

app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+            {
+                foreach ($cities as $city)
+                {
+                    $city->created_by = null;
+                    $city->user_id_last_uploaded = null;
+                    $city->save();
+                }
+            }
+
+            $littercoin = Littercoin::where('user_id', $userId)->get();
```
GeniJaho, line 174: This can be just `Littercoin::where('user_id', $userId)->delete();`.
You can do similar simplifications to the other parts below.

app/Http/Controllers/API/DeleteAccountController.php
```diff
@@ -0,0 +1,323 @@
...
+            $teamUsers = DB::table('team_user')
+                ->where('user_id', $userId)
+                ->get();
+
+            foreach ($teamUsers as $teamUser)
+            {
+                $teamUser->delete();
+            }
+
+            $teams = Team::where('leader', $userId)
```
GeniJaho, line 246: If the team has no other member we delete the team. If the team has other members, we assign the first one as the leader.
Also, I think we need to split where and orWhere for leader and created_by and do different actions for them.

GeniJaho (changes requested):

xlcrr:
Thanks for the review @GeniJaho  - I updated the scripts per location, nice find 

You are probably right about the count checking but I decided to leave the count check there for now

PR #642, opened 2024-01-01
Upgrade PHP

GeniJaho:


xlcrr:
Thanks @GeniJaho im trying to review the PR but its very large! A lot of moving parts makes it hard to test. Might start the php upgrade again and keep it simple

PR #652, opened 2024-09-02
Fix typos

omahs:
Fix typos


xlcrr:
Thank you @omahs ! Will include this in the next release

PR #712, opened 2026-06-27
run specs on all branches

afomi:


.github/workflows/laravel.yml
```diff
@@ -2,13 +2,7 @@ name: Laravel CI
 
 on:
   push:
-    branches:
-      - master
-      - upgrade/tagging-2025
```
xlcrr, line 7: I think we want to keep master but `upgrade/tagging-2025` can be removed

xlcrr (commented):
sorry forgot to submit this review 3w ago, thought it was just a comment!

PR #717, opened 2026-08-14
Fix post-migration tag counts

xlcrr:
Some tags were not correctly migrated during the v5 update

Introducing a fix to track problematic tags and upgrade them 1 by 1


xlcrr:
Prompt hell.

